Issue #548, opened 2013-07-08
Need a way to find out from FlywayException which SQL command failed


bpg-dev:
This feature request partially duplicates #234 bug. 

I would like to have ability to obtain some details regarding failed SQL statements from the generated FlywayException.
For example, the code:

``` java
try{
    Flyway flyway = new Flyway();
    flyway.setDataSource(ds);
    flyway.migrate();
} catch (FlywayException ex) {
    // process ex;
}
```

produces exception like:

```
Caused by: com.googlecode.flyway.core.api.FlywayException: Migration of schema "DEV" to version 3.0.0.001 failed! Please restore backups and roll back database and code!
    at com.googlecode.flyway.core.command.DbMigrate.migrate(DbMigrate.java:199)
    at com.googlecode.flyway.core.Flyway$1.execute(Flyway.java:866)
    at com.googlecode.flyway.core.Flyway$1.execute(Flyway.java:819)
...
```

But I want to tell end user (in my case it is a system administrator) what exactly happened during the migration and give her ability to fix the situation and repeat the attempt.
The current implementation gives only one way to know about the errors -- the log file (tested with Oracle DB):

```
1775 [main] ERROR com.googlecode.flyway.core.command.DbMigrate  - com.googlecode.flyway.core.api.FlywayException: Error executing statement at line 17: CREATE GLOBAL TEMPORARY TABLE "TEMP_XXX_INSERTED" (
  "ID" INTEGER NOT NULL,
  CONSTRAINT "TMP_XXX_INSRTD" PRIMARY KEY ("ID")
) ON COMMIT DELETE ROWS>

1775 [main] ERROR com.googlecode.flyway.core.command.DbMigrate  - Caused by java.sql.SQLException: ORA-00922: missing or invalid option
```

I would like to see an extension of FlywayException interface (or subclass e.g. FlywayMigrationException) that can give me a list of errors occurred during migration (or only the first error -- because Flyway does not apply subsequent migrations), so it can be processed by business logic.

PR #554, opened 2013-07-14
Implementation of #548 feature request

bpg-dev:
Added MigrationResult class for capturing of JDBC exception during
migration if any, and wrap it into FlywayException.


axelfontaine:
@bpg Could you please get in touch with me via email? (axel at boxfuse.com) Thanks!

Issue #563, opened 2013-07-24
MigrationVersion compareTo is very slow


pwielgolaski:
I noticed during profiling our app that migration takes(4s - 4 migrations) quite long and after tracking down I noticed that implementation of MigrationVersion is quite ineffective as it don't cache pattern and use patterns to parse string every time compareTo is called, which is often as you put and get it from Map.

PR #564, opened 2013-07-24
performance improvement for MigrationVersion.compareTo 

pwielgolaski:
issue described in https://github.com/flyway/flyway/issues/563
in general moved logic of parsing to construction phase and make some reusing of
cached patterns

result of micro-brenchmarking are promising compareTo ~ 33% of old implementation where construction is faster but not so significant.


axelfontaine:
@pwielgolaski Could you get in touch with me via email (axel at boxfuse.com)? Thanks!

PR #601, opened 2013-09-12
Feature placeholder optional strict enforcement

bigslack:
I have added the property `flyway.placeholderStrict` which tells flyway to enforce the unmatched placeholders exception or not.

We have several situations where we want to replace some placeholders but not others, this can change depending on the test environment we set up (sometimes other systems change the placeholders upstream or down stream). I saw a few other posts on unmatched placeholder exceptions in the forums and thought this may be a good approach to addressing the issue.

Note that the default is set to false which is not the current default behavior of Flyway but I think it fits better with the most general use case.

Look forward to your thoughts.


axelfontaine:
Closing as part of a cleanup of old issues.

Issue #562, opened 2013-07-24
Cygwin support


KevinSheedy:
From a cygwin shell on Windows XP, type `flyway migrate` and I get the following error message:

```
$ flyway migrate
Error: Could not find or load main class com.googlecode.flyway.commandline.Main
```

PR #603, opened 2013-09-13
fix problem of running command line under cygwin 

pwielgolaski:
cygwin is complaining about paths and it requires . as current path
additionally I fixed problem when JAVA_HOME has some spaces.

I made also conditional call of tput so it does not complain on cygwin when it is missing

It solves issue https://github.com/flyway/flyway/issues/562


axelfontaine:
@pwielgolaski Did you receive my previous message? Could you get in touch with me via email? (axel at boxfuse.com) Thanks!

PR #664, opened 2013-12-13
Added check to remove versoning on DB2 tables during clean.

boothen:
Versioning tables are automatically deleted when the table they version are deleted during a clean.

Dropping versioning before the dropping of the tables allows the tables to be dropped without issue.


axelfontaine:
@boothen Could you get in touch with me via email? (axel at boxfuse.com) Thanks!

Issue #687, opened 2014-01-31
Oracle cleanup - flyway fails on tables that have Flashback enabled


awal11:
I have a schema with tables set to use Oracle Flashback/Total Recall (full audit implemented completely on database side). 

After I create the tables with command like 

``` oracle
ALTER TABLE USERS FLASHBACK ARCHIVE FDA_TRAC
```

Oracle creates 'magic' tables like SYS_FBA_HIST_61527 that **flyway fails to delete automatically**, throwing: 
ORA-55622: DML, ALTER and CREATE UNIQUE INDEX operations are not allowed on table "SCOTT"."SYS_FBA_HIST_61527"

To be able to clean the schema one would need to execute 

``` oracle
ALTER TABLE USERS NO FLASHBACK ARCHIVE
```

before attempting to delete anything.

So how should I help:
- if this pull request https://github.com/flyway/flyway/pull/629 gets accepted - I guess I don't need to do anything - is there a chance for it?
- I could update  OracleSchema.java to detect and 'NO FLASHBACK' appropriate tables automatically - would such patch be accepted?
- if so - should I add a test case for it? It will end up @Ingored as Flashback not enabled in most oracle instances and not very popular.

PR #689, opened 2014-02-03
Fixing https://github.com/flyway/flyway/issues/687 - added code that can remove Flashback tables in Oracle and a test case for it. Also fixed one incorrect Oracle test configuration method.

awal11:


axelfontaine:
@awal11 Could you get in touch with me via email? (axel at boxfuse.com) Thanks!

Issue #651, opened 2013-11-27
Pre and post hooks for each operation


npeters:
Hello,

I would like to have Hook (Listener) on pre-migration and post-migration.
It can be used to stop and start replication, move db role or do some other stuff

PR #709, opened 2014-03-03
PR for issue 651

dlbunker:
... hooks on the db migrate task.  This is an effort to address issue 651 located here:

https://github.com/flyway/flyway/issues/651

This task could be updated to include a context that gets passed from listener to listener if there is a need for that.  This would allow listeners to communicate with each other.


axelfontaine:
@dlbunker Could you please contact me by email at axel at boxfuse dot com ? Thanks!

PR #765, opened 2014-05-26
 flyway#754 fix bug on quote together with comment

nivolg:
This bug is common for all databases. It occurs when quote follows the comment start or precedes the comment end without whitespaces.
I add sql statement for testing all databases into common V1__Comment.sql script.
A also add V1__Comment.sql to dbsupport/oracle/comment for testing single line comment. It cannot be put into common file because mysql syntax is not supported single line comments without whitspace after --

My pull requet #764 is wrong. This is correct.


axelfontaine:
@nivolg Could you please get in touch with me via email? (axel at boxfuse.com) Thanks!

PR #778, opened 2014-06-17
Maven plugin: Expose new DB version in "flyway.current" property after migration

michaelyaakoby:
Hi,

I'm using flyway in my code and maven and its awesome.
Recently I needed to be able to get, in maven, the current version after migration - my build is automatically invoking the migrate goal during install.

So I've extended the maven-plugin so after migration it records the current DB version in the "flyway.current" property.

In order to test this, I've also extended the maven-plugin-largetest to test the "executions" section of the plugin.

Right now I'm using a local version that includes this patch - will be great if you can merge it into flyway.

Thanks,
M.

BTW, this is my first pull request... so please educate me if I'm doing it wrong :-)


axelfontaine:
@michaelyaakoby Could you get in touch with me via email? (axel at boxfuse.com) Thanks!

PR #781, opened 2014-06-19
Support for migrating multiple schemas at once

jyore:
Current codebase did not support migrating multiple db's in a single maven command.  The use case for this is multi-tenant systems that have data isolation based on the database schema. Running maven once per tenant is not feasible. 

The implementation provided may need some work, but I wanted to keep the current implementation in-tact and allow for the option to run migrations over the entire flyway.schemas list. 

To accomplish this, I added a property (flyway.multipleDbMode), defaulted to false.  When true, this will migrate every schema in the list provided to flyway.schemas.  When false, this will migrate just the first of the list, as was in place in the current codebase. Both the gradle and maven plugins have been updated to set the property appropriately as well.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
@@ -16,12 +16,26 @@
 package org.flywaydb.core;
 
 
+import java.sql.Connection;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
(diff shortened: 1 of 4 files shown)


axelfontaine:
Any news?

axelfontaine:
Closing due to lack of response from original poster

alayor:
What's the workaround for Gradle?
I really need this feature.

juliahayward:
See also #2995

PR #817, opened 2014-08-08
Transaction "switch"

soldierkam:
Since Postgresql does not support "ALTER TYPE ... ADD VALUE ..." inside transaction its required to disable transaction for some scripts. This commit allow to add comment "--not-transactional" at begin of script to disable transaction.


axelfontaine:
Closing due to lack of response from original poster

PR #828, opened 2014-08-26
Firebird support

tweck:
First shot at [Firebird](http://www.firebirdsql.org/) support.
MigrationTestCase succeeds :white_check_mark:

the only crude part: flyway needs to run in some kind of auto-commit mode, since Firebird does not allow the mixing of DDL & DML within a single transaction - which would be nothing new as other DBs like Oracle also don't support it, but the problem here is that Firebird does also not perform implicit commits after DDL operations, so flyway could not create its metadataTable and insert the initial version otherwise.

maybe a better approach would be to include explicit commit-statements in the sql-scripts, but as this feels like a more heavy impact on the code base which is required for just this one DB, I just took the pragmatic route with a specialized JdbcTemplate.

### DIFF
--- a/flyway-core/pom.xml
+++ b/flyway-core/pom.xml
@@ -122,6 +122,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.firebirdsql.jdbc</groupId>
+            <artifactId>jaybird-jdk16</artifactId>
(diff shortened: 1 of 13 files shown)


axelfontaine:
Any news?

axelfontaine:
Closing due to lack of response.

PR #879, opened 2014-11-13
Patch to submit flyway plugin to Gradle Plugin Portal

aalmiray:
Please see additional instrunctions laid out at http://plugins.gradle.org/submit
Update suggested values at gradle.properties and publishing.gradle at your leisure


axelfontaine:
This has now been done through bintray and their automatic sync to both Maven Central & the Gradle Plugin portal.

PR #903, opened 2014-12-03
Fix an error where MigrationVersion wouldn't parse correctly

andrewlmurray:
Presently MigrationVersion parses to a List of Longs internally. If any of the pieces of a version string overflow a long it will fail to parse with a misleading error message. This is problematic for two reasons:
1. It can be useful to have version numbers longer than longs.
2. The error message provided to the user is incorrect in the case of overflow and is confusing.

Take for example a migration with the name:

```
V20141203182115912052_00001__create_user.sql
```

(Here we're using the date/time from a shared ntp server as the prefix of a version number, which overflows a long) upon being parsed in trunk we'd get this error:

```
org.flywaydb.core.api.FlywayException: Invalid version containing non-numeric characters. Only 0..9 and . are allowed. Invalid version: 20141203182115912052.00001
```

Which is obviously incorrect. 

This patch changes the internal storage of the migration version to use a BigInteger instead of a Long, thus allowing version sections to be of arbitrary length.  I ran this with the non-commercial test suite and everything passed, but I haven't run with the commercial stuff.

### DIFF
--- a/flyway-core/src/test/java/org/flywaydb/core/api/MigrationVersionSmallTest.java
+++ b/flyway-core/src/test/java/org/flywaydb/core/api/MigrationVersionSmallTest.java
@@ -164,5 +164,12 @@ public class MigrationVersionSmallTest {
     public void alphaNumeric() {
         MigrationVersion.fromVersion("1.2.1a-3");
     }
+
+    @Test
+    public void testWouldOverflowLong() {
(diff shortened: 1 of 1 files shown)


axelfontaine:
@andrewlmurray Could you please get in touch with me via email (axel at boxfuse.com)? Thanks!

axelfontaine:
@andrewlmurray Did you get my previous message? Could you please get in touch with me via email (axel at boxfuse.com)? Thanks!

PR #911, opened 2014-12-16
Update start scripts

mlliso:
Enable flyway to use CP environment variable to point additional classpath locations.


axelfontaine:
Superseded by #966

Issue #906, opened 2014-12-08
Option to validate only until current version of the schema


sambekar:
I want to validate my migration scripts against an existing database on which I have applied some migration. Now my classpath contains scripts which would be applied on the said database in future.
After 3.0 the validate method defaults the option to allow pendingOrFuture migration in validate method to false.

Whats the purpose of having this configurable item if its defaulted to false? 
Can we make it configurable?

doValidate(connectionMetaDataTable, connectionUserObjects, migrationResolver, metaDataTable, schemas,false);

PR #912, opened 2014-12-23
Issue #906: Made validating against an existing database with pending or future migr...

sanjayd:
...ations configurable


axelfontaine:
Superseded by #913

Issue #918, opened 2015-01-07
Explicitly drop Views then doing a Flyway.clean() on a PostgreSQL db


gergelyszakacs:
Database Views are not dropped when a Flyway.clean() is issued. If the migration scripts baseline the database with some existing tables that cannot be touched and just adds some Views (probably in a different schema than the tables themselves) then the clean() call will leave those Views in place. So a new feature is needed that explicitly drops these Views.

axelfontaine:
I am trying to understand your exact scenario.

DB with existing Schema A with table A1
Flyway.schemas set to B
Flyway creates schema B and a migration adds a view B1 that maps to A.A1
Flyway.clean (schema B only) does not drop B1

Is this correct?

gergelyszakacs:
Hi Axel,

Thanks for getting back to me so soon. Yes, this is my scenario.


--
Gergő

logicbomb:
:+1: the fix as implemented by @gergelyszakacs  resolves the issue

teicher:
Hi Axel, looks to me like this will not work for materiaized views. (need to call DROP MATERIALIZED VIEW ...)
Cheers, Tom.

axelfontaine:
@teicher Please file a new issue and include a script that Flyway fails to clean.

teicher:
Okay Okay, with 3.2.1 it works even for materialized views. But not due to 1dae524, I'd say ;-)

teicher:
Okay, so the dependency between views (and materialized ones) is the problem:
https://github.com/flyway/flyway/issues/1129

PR #919, opened 2015-01-07
#918 : Implementing explicit VIEW drop for PostgreSQL

gergelyszakacs:
- Created the abstract View class (a slight modification of Table)
- Created the PostgreSQLView class as an implementation
- Modified the Schema class to have View[] allViews() and View[] doAllViews()
- Modified the PostgreSQLSchema class to implement View[] doAllViews()
- Modified the rest of the vendor specific Schema classes to have a dummy impl

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/dbsupport/Schema.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/dbsupport/Schema.java
@@ -180,6 +180,27 @@ public abstract class Schema<S extends DbSupport> {
     protected abstract Table[] doAllTables() throws SQLException;
 
     /**
+     * Retrieves all the views in this schema.
+     *
+     * @return All views in the schema.
+     */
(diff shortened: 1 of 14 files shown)


axelfontaine:
Thanks for the pull request.

I currently see two issues here:
- lots of unnecessary reformating
- the view abstraction

Could you resubmit it with introducing both of the above?

Thanks
Axel

gergelyszakacs:
Hi Axel,

Reformatting: no problem, I was using Eclipse. I install ItelliJ IDEA and
redo without formatting.
View abstraction: I'm not sure I understand what's wrong with it.


--
Gergő

axelfontaine:
I'm fine fine with the view abstraction, but then all implementing classes should correctly be changed to use it (shouldn't be that much work though).

Looking forward to the update!
Axel

axelfontaine:
Superseded by most recent commit. Closing.

PR #922, opened 2015-01-09
Escape command-line arguments.

marcenuc:
Following the instructions at
http://flywaydb.org/documentation/existing.html on Linux, the following
command fails with `ERROR: Invalid operation: version`:

    ./flyway baseline -Dflyway.baselineVersion=1 \
        -Dflyway.baselineDescription="Base version"

The problem is that the `flyway` Bash script does not properly escape
the command line arguments.

This patch adds quotes around all variables to fix this error. As a
side effect, it also allows to have spaces in the Flyway installation
directory name.

### DIFF
--- a/flyway-commandline/src/main/assembly/flyway
+++ b/flyway-commandline/src/main/assembly/flyway
@@ -16,9 +16,9 @@
 #
 
 # Save current directory
-OLDDIR=`pwd`
+OLDDIR="$(pwd)"
 cygwin=false
-case "`uname`" in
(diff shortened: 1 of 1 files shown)


axelfontaine:
Thanks. Flyway already works when installed in directories with spaces (we explicitly test for that).

However, nice catch with the arguments! I haven't merged the PR as it contained unnecessary changes, but I added the quoting around the one argument.

Cheers
Axel

PR #924, opened 2015-01-09
Fixed #917 Adding flywayCurrentVersion task

### LINKED ISSUE #917: After running flywayMigrate in SBT there is no straightforward way to know current db version
We migrate from MVN to SBT. One of the features that we used in MVN plugin was `${flyway.current}` that we used to write inside the *MANIFEST.MF* file (and used it for support info in runtime). After reading the SBT plugin doc I was unable to find an easy way to get the current db version, so it could be nice-to-have task that returns it.

j-unior:


axelfontaine:
I am unsure about adding another task. Is there no way to define a property dynamically as part of flywayMigrate?

j-unior:
Why wouldn't flywayMigrate return the current version instead of `Unit` (or even a case class with more detailed information regarding the migrations)?
Optionally we can make a `flywayCurrentVersion` setting being `Option[String]` and initialize it after migration, yet it feels wrong and not functional.
Still having a separate tasks makes some sense to me (though I understand it's not the best performance-wise). What if the user wants to evaluate the current version before migration for some kind of additional build logic?

axelfontaine:
If it is possible for flywayMigrate to return the version, I would favor that option.

@cdedie Thoughts?

axelfontaine:
Closing as I strongly favor having flywayMigrate return the version instead

Issue #929, opened 2015-01-14
Apache Phoenix Support


jmahonin:
Phoenix (http://phoenix.apache.org/) is an open source relational DB over HBase that exposes a JDBC driver.

I've got a working implementation as well as unit tests that I'll be putting in a pull request. It works against a stand-alone Phoenix cluster, but the tests run against an in-memory HBase cluster. Some comments from the commit message below:

It has been set up as an EmbeddedDB, although the Phoenix tests
should be disabled by default unless run in the new 'PhoenixDBTest'
profile.

Notable changes to core classes:

* MetaDataTableImpl:

Due to Phoenix's peculiar syntax, the hardcoded statements in
addAppliedMigration() and updateChecksum() won't work. It now
attempts to load a .sql template and run that, if it exists.

* MigrationTestCase:

Added checks for applied migrations of type MigrationType.SCHEMA
and adjusted asserts accordingly. Also added a 'getBaseDir()'
method, and replaced direct calls to BASEDIR with this. Several of
the tests in MigrationTestCase can then be used in the Phoenix
tests by overriding that method.

anirudha13:
@jmahonin I have ulterior motives in seeing this work merged with the mainline. Please let me know how I can help.

jmahonin:
@anirudha13 Me too!

If you could do some quick testing with it and let me know how it goes, that'd be appreciated. We're using it internally already, and it passes the migration test suite, but having some assurance from another party would be appreciated.

anirudha13:
@jmahonin ran some very basic tests using the flyway command line utility from this pull request. Things work like a charm. Will keep you updated if something breaks.

jmahonin:
Fantastic. Out of curious its, what version of Phoenix were you using?

jmahonin:
*sigh* On mobile, excuse the typos. "Out of curiosity" :)

anirudha13:
v4.3.1 with HBase - 0.98.6 CDH 5.3.2

chiastic-security:
Good work, guys!

Is this likely to make it into a release in the near future? I'm wanting to use flyway with Phoenix, and not sure whether to wait for it to be pulled in, or find another way.

anirudha13:
@jmahonin while running the unit tests I get this failure, 

```
Running org.flywaydb.core.internal.dbsupport.phoenix.PhoenixMigrationMediumTest
[...]

PR #930, opened 2015-01-14
Add support for Apache Phoenix (http://phoenix.apache.org/)

jmahonin:
PR for issue https://github.com/flyway/flyway/issues/929

It has been set up as an EmbeddedDB, although the Phoenix tests
should be disabled by default unless run in the new 'PhoenixDBTest'
profile.

Notable changes to core classes:

* MetaDataTableImpl:

Due to Phoenix's peculiar syntax, the hardcoded statements in
addAppliedMigration() and updateChecksum() won't work. It now
attempts to load a .sql template and run that, if it exists.

* MigrationTestCase:

Added checks for applied migrations of type MigrationType.SCHEMA
and adjusted asserts accordingly. Also added a 'getBaseDir()'
method, and replaced direct calls to BASEDIR with this. Several of
the tests in MigrationTestCase can then be used in the Phoenix
tests by overriding that method.

### DIFF
--- a/flyway-core/pom.xml
+++ b/flyway-core/pom.xml
@@ -140,6 +140,34 @@
             <artifactId>jtds</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.hbase</groupId>
+            <artifactId>hbase-it</artifactId>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hbase</groupId>
+            <artifactId>hbase-testing-util</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hbase</groupId>
+            <artifactId>hbase-hadoop-compat</artifactId>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hbase</groupId>
+            <artifactId>hbase-hadoop2-compat</artifactId>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.phoenix</groupId>
+            <artifactId>phoenix-core</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
@@ -245,7 +273,8 @@
                     <plugin>
                         <artifactId>maven-failsafe-plugin</artifactId>
                         <configuration>
-                            <excludedGroups>org.flywaydb.core.DbCategory$InstallableDB</excludedGroups>
+                            <!-- Don't run the embedded Phoenix test by default, it has its own profile -->
+                            <excludedGroups>org.flywaydb.core.DbCategory$InstallableDB,org.flywaydb.core.DbCategory$Phoenix</excludedGroups>
                         </configuration>
                     </plugin>
                 </plugins>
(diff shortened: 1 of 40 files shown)


jmahonin:
Running the same command as Travis CI works on my machine:
mvn install -P-CommercialDBTest -DskipTests=true -B

Also the Phoenix tests can be run stand-alone as:
mvn install -P PhoenixDBTest -P-CommercialDBTest -P-InstallableDBTest

They take around 5 minutes or so on my machine. There's a zookeeper exception that's benign, and appears to be normal behaviour with the HBaseTestingUtility

jmahonin:
Just pinging for an update. I'm open to any criticisms and concerns you may have with this PR.

I don't have any issues building on my machine, it appears the Travis CI issue may have been network related:

```
[ERROR] Failed to execute goal com.mycila.maven-license-plugin:maven-license-plugin:1.9.0:format (default) on project flyway-parent: Execution default of goal com.mycila.maven-license-plugin:maven-license-plugin:1.9.0:format failed: Plugin com.mycila.maven-license-plugin:maven-license-plugin:1.9.0 or one of its dependencies could not be resolved: Could not transfer artifact log4j:log4j:jar:1.2.12 from/to central (https://repo.maven.apache.org/maven2): GET request of: log4j/log4j/1.2.12/log4j-1.2.12.jar from central failed: Connection reset -> [Help 1]
```

axelfontaine:
Hi Josh,

sorry for the late reply. It's a very busy time for me right now. I'll get back to you in the next 2-3 weeks and we'll look at integrating it in Flyway 4.0.

Cheers
Axel

jmahonin:
No problem, thanks for the update.

jmahonin:
Just a heads up, I'm going to re-work this to support Apache Phoenix 4.3.0, and re-work the specific catalog queries into using the DatabaseMetaData interface instead, which should help with forward and backward compatibility.

If you did have any suggestions on a better way to handle the custom updateChecksum and updateMetaDataTable calls, I'd love to hear them. Also, I've made changes to MigrationTestCase to be a bit more generalized, but if you have suggestions there as well please let me know.

axelfontaine:
Hi Josh,

thanks for the update. Why did you need to duplicate all those test cases? Also why did you need all those placeholders for the column names? (just trying to understand)

Would you also be willing to
- maintain this in the future and answer an user question/solve any issues when they come up?
- provide a pull request against flywaydb.org (I'll create a flyway 4.0 branch)?

Cheers
Axel

jmahonin:
Hi Axel,

Unfortunately Phoenix has some peculiar syntax. INSERTs are UPSERTs, and the CREATE TABLE syntax and datatypes don't work with a lot of the canned SQL in flyway.

I duplicated any test case where specific SQL was being executed from a test case, or loaded loaded from a file that wasn't valid Phoenix SQL. Unfortunately, that was most of them. You'll see the test SQL files are copied as well in test/resources/dbsupport/phoenix/sql.

Another option would be to update MigrationTestCase so that the SQL location folder ("migration"), can be overridden through a subclass to a different location, e.g. "migration/dbsupport/phoenix". I did it a bit with the 'getBaseDir()' call, but I wasn't thorough with it.

I'm willing to continue to maintain this and can provide a PR for documentation against flywaydb in line with the other supported DBs.

Thanks!

Josh

jmahonin:
Oh, re: placeholders, you're right, I can hardcode in the column names in the updateMetaDataTable.sql and updateChecksum.sql. 

I think my reasoning behind that was the column names were in the code already there anyway, so if they ever changed, at least they'd only change in one file. I'm good either way, your call.

axelfontaine:
Hardcode them. I'm a fan of keeping this simple.

jmahonin:
Agree.

Any opinions on adding something like a 'getMigrationDir()' in MigrationTestCase which would return "migration", and subclasses could override it with their own dbsupport locations? e.g. "migration/dbsupport/phoenix/sql"? 

The flyway.setLocations() would need to be updated to use the 'getMigrationDir()', but I think that would let me remove just about every one of those duplicated unit tests.

jmahonin:
Updated this PR to:
https://github.com/flyway/flyway/pull/975

PR #935, opened 2015-01-18
Print error messages to stderr(System.err) from CLI

karl82:
I need to redirect error messages from command line into separate file. It's possible to use ```|grep 'ERROR:'``` but it will not catch exception stacktraces, because they are not prefixed with ```ERROR:```

I think this is a good behavior for command line tool

### DIFF
--- a/flyway-commandline/pom.xml
+++ b/flyway-commandline/pom.xml
@@ -54,6 +54,11 @@
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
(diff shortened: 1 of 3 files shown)


axelfontaine:
Thanks Karl! This makes a lot of sense. I'll merge it for 4.0 (due in April) as it isn't backwards compatible.

Cheers
Axel

karl82:
great! thx

axelfontaine:
Merged. Thanks once again!

karl82:
awesome!

axelfontaine:
@karl82 Could you please get in touch with me via email? (axel at boxfuse.com) Thanks!

PR #969, opened 2015-03-11
Fix SQL syntax error in MetaDataTableImpl.java

avesse:
Syntax error encountered on MySQL due to missing space after "OR" in SQL query in MetaDataTableImpl.java.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/metadatatable/MetaDataTableImpl.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/metadatatable/MetaDataTableImpl.java
@@ -319,7 +319,7 @@ public class MetaDataTableImpl implements MetaDataTable {
 
         try {
             int count = jdbcTemplate.queryForInt(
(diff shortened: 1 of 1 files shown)


axelfontaine:
Thanks for the pull request. What version of MySQL is exhibiting this behavior? (Our MySQL tests don't seem affected, so I'm wondering)

avesse:
It's actually not MySQL, but Clustrix, which uses MySQL's syntax.

axelfontaine:
Thanks! Merged. I've also listed you on the hall of fame page. Can you tell me a bit more about your experiences with using Flyway and Clustrix together? (you are the first one I'm aware of, so this is just personal curiosity)

avesse:
Cool :)

So far it's been good; we've only just started using it on Clustrix, but to date we've set up a new db using flyway, ran a migration on it, as well as a baseline plus migration on an existing db. This issue is the only one we've encountered.

All our other tooling also only supports MySQL explicitly, not Clustrix, but for the most part it all just works.

PR #985, opened 2015-04-08
Log the update count with DEBUG priority after running a SQL script

thelateperseus:
Our DBA team want to be able to see the number of rows updated by Flyway.  We occasionally have data migrations via an UPDATE statement, and we want to record the number of rows updated in each database environment for auditing purposes.  I've logged with DEBUG priority because most users obviously don't need this feature.


axelfontaine:
Thanks! Makes sense. Merged.

PR #1006, opened 2015-04-30
fixes #178

### LINKED ISSUE #178: Support for reference-partitioned tables (Oracle 11g)
_Original author: sandro.d...@gmail.com (March 28, 2013 09:17:26)_

Hi, 

We introduced reference partitioning (Oracle 11g). As a consequence, the flyway clean goal fails with &quot;ORA-14656: cannot drop the parent of a reference-partitioned table&quot;. 

Flyway drops tables by using &quot;drop table xyz cascade constraints purge&quot;. Unfortunately, this does not work for reference-partitioned tables.

We see two possible ways to solve this issue:
1) Introducing some plsql code (in OracleDbSupport class) for dropping tables in a &quot;bottom to top&quot; order.
2) The clean goal supports a dedicated script directory (like migrate does). These scripts are executed (if any) by the clean goal (replacing the default drop behaviour).

Solution 2 is more like a new feature than a bugfix. Which solution would make sense to you? Or any less invasive ideas for solving this issue? 

If you wish, we could contribute to solution 1 and provide a pull request. We wrote a plsql script dropping the user tables in correct order and could try to integrate it into OracleDbSupport.

Cheers,
Sandro

PS: I attached the plsql script (not the final version, but just to give an idea...)

_Original issue: http://code.google.com/p/flyway/issues/detail?id=465_

pingw33n:


axelfontaine:
Thank you. Can you add a test case as well?

axelfontaine:
Thanks!. Merged.

Issue #253, opened 2013-06-25
Ignore line endings when calculating checksums


flyway:
_Original author: micha...@utest.com (September 28, 2011 15:49:50)_

We ran into the following problem.

Developer on Mac/Linux creates a sql migration file and checks it into Git.  It runs on production (which is on Linux) just fine.

Then a Windows developer takes a copy of the production database and loads it into their local machine.  But when they pulled the sql migration file from Git, it automatically converted the line endings to CRLF (Windows style).  Now Flyway validate and migrate fail saying there is a checksum mismatch.

One solution I can think of would be to ignore line endings when doing the checksum. 

_Original issue: http://code.google.com/p/flyway/issues/detail?id=167_

kylecordes:
I am really surprised to run in to this problem - it is exactly the sort of thing I expected to "Just Work".

I am even more surprised to see resistance to fixing it!

Hopefully the suggest workaround of converting all files to LF-only, before letting Flyway see them, won't be too awful to hack together.

axelfontaine:
Hi Kyle,

you may be surprised and you're probably thinking that since it's only a matter of adding ```script.replace("\r", "")``` it should be fixed in an instant. However there is more than meets the eye. This seemingly trivial fix **invalidates the metadata table on every single Flyway installation out there**. And so great care must be taken to make this transition as smooth as possible, with no surprises. And since Flyway follows the tried and true principles of semantic versioning, it is something that'll have to wait until the next major release.

Cheers
Axel

kylecordes:
While that smallest fix has the problem above, the software is *already* completely broken in the way you describe for anyone working across platforms. So while invalidating metadata is a bad thing, it is already invalidated for lots of people.

There is an easy fix could be made, even without a major version: add an optional setting that brings in the new behavior now. Well I very much agree with the bias against adding settings, a "stop being completely broken on cross platform projects" setting would be welcomed, even by the most vigorous opponent of settings.

There is also a more involved approach, which avoids the setting and produces immediately good results: start calculating the checksum both ways, and accept either is a match. This would be perfectly backward-compatible, so it would not require a major version, and it would fix the problem without invalidating any old metadata.

axelfontaine:
Hi Kyle,

Truth be told the problem is really Git's "convenient" autocrlf feature. However, since it dominates the market, I agree we need to compensate for it within Flyway, so that users have a good experience.

[...]

PR #1031, opened 2015-06-09
Fix for issue #253 Ignore line endings when calculating checksums

Megaprog:
Calculation migration checksum by ignoring CR symbols (\r) and BOM sequence for UTF-8 encoding. If there are no such symbols in migration checksum will not changed.

I created two .txt files for testing checksums but can't pull CRLF symbols to first one (BOM sequence was pulled). Who wants can do it youself.

#253


axelfontaine:
Closing has this has been made obsolete by the implementation in master.

Issue #1043, opened 2015-07-02
Migration with Flyway.setPlaceholderReplacement(false) fails with "org.flywaydb.core.api.FlywayException: No value provided for placeholder expressions"


lukosan:
The fix for #945 causes an exception to be thrown when running Flyway.migrate() because the method in PlaceholderReplacer which checks whether there are unmatched placeholders (checkForUnmatchedPlaceholderExpression) will match every char in the String, resulting in:

Exception in thread "main" org.flywaydb.core.api.FlywayException: No value provided for placeholder expressions: 	,  , !, ", #, $, %, &, ', (, ), *, +, ,, -, ., /, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, :, ;, <, =, >, ?, @, A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P, Q, R, S, T, U, V, W, X, Y, Z, [, ], _, a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p, q, r, s, t, u, v, w, x, y, z, {, |, }, ~.  Check your configuration!

axelfontaine:
Thank you Luke! I ended fixing this slightly differently than your PR, but it wouldn't have been this easy hadn't you provided the testcase and done the investigation.

Cheers
Axel

PR #1045, opened 2015-07-02
#1043 checkForUnmatchedPlaceholderExpression now only performs check …

lukosan:
…if both prefix are suffix are not empty

This is a fix for  #1043


axelfontaine:
Thank you Luke! I ended fixing this slightly differently than your PR, but it wouldn't have been this easy hadn't you provided the testcase and done the investigation.

Cheers
Axel

Issue #1039, opened 2015-06-29
Support for migrating multiple identical schemas at once


CSchulz:
I would like to see something like #781 and http://stackoverflow.com/questions/15476659/using-flyway-to-patch-multiple-identical-schemas/.
I think the easiest implemtation would adding multiple Maven Plugin executions suggested by Alex.

Add another parameter which allows to switch the working modes when passing multiple schema.

rodgilbert:
Would like this capability for Gradle as well.

rodgilbert:
This is not a solution for Gradle users.   No work around.

Multiple identical schemas

If you have multiple identical schemas, say one per tenant, invoke Flyway in a loop and change flyway.schemas to match the name of the schema of the current tenant.

brian-villanueva:
Invoking Flyway in a loop works, but if you have multiple identical schemas (one per tenant), why wouldn't you want to be able to run the migrations in parallel? We're finding that with many tenants, running migrations in a loop can take a very long time. Has anyone implemented a multithreaded migration for multiple schemas?

CSchulz:
Did you tried my PR? Doesn't decrease it the execution time?
You haven't the overhead of invoking multiple times Flyway.

brian-villanueva:
@CSchulz I have not tried it, but it looks like the PR has not yet been accepted. We are not likely to implement an unsupported Flyway version, though I like your idea!

juliahayward:
See also #2995

PR #1048, opened 2015-07-06
Support for migrating multiple schemas at once

CSchulz:
First and simple implementation for #1039


axelfontaine:
Closing, because if we chose to implement this, it will most certainly be implemented in the core and not in a specific plugin.

Issue #724, opened 2014-03-29
Prompt for userId and password in commandline mode


sachinwalia2k8:
I think it is difficult to implement userId and password encryption in flyway.conf file without making other compromises such as exposing encryption keys. How about if in commandline mode flyway prompts user for database credentials? That would still be very secure compared to storing in clear text in conf file.

PredatorVI:
Pull request for fix submitted.  I'm new to the GitHub process.  Hopefully it gets accepted and released soon.

axelfontaine:
This has now been fixed. Prompting will happen automatically if either user or password have not been supplied via any other means.

Cheers
Axel

smanoharaw:
Hi axelfontaine,
I love the product. I tested it yesterday on my Oracle db. very easy to use. But I really need the passwordPrompt functionality? Where can I get the flyway version with this? Or, if I could get the code I can create the jar file? Please let me know how to go about this. We are going live in a month or so. 
Thanks

szxnyc:
Hello,
I'm using version 4.0 and didn't supply a password but not getting prompted. Anything I'm doing wrong here?

**flyway.conf**
`#flyway.password=`

```
[szxnyc@orbiter scripts]$ flyway info
Flyway 4.0 by Boxfuse

ERROR: Unable to obtain Jdbc connection from DataSource (jdbc:mysql://localhost:3306/mydb) for user 'devuser': Could not connect: Access denied for user 'devuser'@'localhost' (using password: NO)
```

PredatorVI:
I downloaded the latest client as well and verified this is not working.  I pulled the latest code in order to debug it but can't seem to build it due to missing dependencies:

**Building flyway-core:**
`Failed to execute goal on project flyway-core: Could not resolve dependencies for project org.flywaydb:flyway-core:jar:0-SNAPSHOT: The following artifacts could not be resolved: sqlline:sqlline:jar:1.1.8, com.sap.db.jdbc:ngdbc:jar:1.103.0: Failure to find sqlline:sqlline:jar:1.1.8 in mynexus`

The only version of sqlline available in maven central is version 1.1.9.

**Building flyway-commandline:**
[...]

PR #1051, opened 2015-07-08
Fix for #724:  Added command-line flags (-userPrompt, -passwordPrompt) to prompt for…

PredatorVI:
… the user and/or password.

https://github.com/flyway/flyway/issues/724

This issue was opened last year, but as I have recently recommended and begun implementation of FlywayDB in my current company, our OPS department really wanted this feature for production so passwords don't appear in the shell history.

I've added a first implementation attempt.  I think it matches coding standards required.  I've tested it with the current v3.2.1 core library.

It uses the java.io.Console.readPassword() method which masks input from prying eyes.


axelfontaine:
Thank you for the PR. It was very helpful in implementing the fix! I've opted for a slightly different solution without the additional config parameters (see comment in #724)

Cheers
Axel

Issue #1059, opened 2015-07-20
Upgrade Spring JDBC dependency to the latest version


guylabs:
Hi,

the Spring JDBC version used in Flyway is 2.5.6 and the goal would be to update it to the latest Spring version.

Thanks and regards,

Guy

axelfontaine:
Closing. See my comment on pull request.

PR #1060, opened 2015-07-20
Upgrade Spring JDBC dependency to the latest version

guylabs:
Pull request for #1059.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -97,7 +97,7 @@
     </repositories>
 
     <properties>
-        <version.spring>2.5.6</version.spring>
+        <version.spring>4.1.7.RELEASE</version.spring>
         <version.appengine>1.9.10</version.appengine>
         <version.maven>2.2.1</version.maven>
         <version.gradle>1.6</version.gradle>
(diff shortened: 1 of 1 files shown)


guylabs:
Hi

Well I think it would be good to at least be on the same major version as the official Spring Boot integration mentioned here: https://docs.spring.io/spring-boot/docs/current/reference/html/howto-database-initialization.html#howto-execute-flyway-database-migrations-on-startup

When I have a look at the project and I see that Flyway is still using Spring `2.5.6` (which is nearly 7 years old) and has Spring Boot support then this somehow does not fit.

And the other point is that for you it would not change anything as you only use the `JdbcTemplate` for the Spring migrations, so there is no additional work or testing needed.

But at the end it is up to you to update it or not.

Thanks anyway to have a look at it.

Regards,

Guy

axelfontaine:
Well the dependency is optional and there is no runtime conflict. So unless I missed something, I do not see the issue.

guylabs:
Yes that is true, it is just cosmetics. ;) But I think it just makes a better look to have it updated in the pom, that is all.

Issue #1062, opened 2015-07-22
API: Make it possible to use pre-instantiated Java-based migrations


guylabs:
Hi,

as we are using Spring with Flyway we want to use other beans inside the `SpringJdbcMigration`s. Therefore we created a new resolver which retrieves the `SpringJdbcMigration`s from the Spring application context. This way we are able to define the migrations inside the application context and reference other beans we would need for this migration inside the `SpringJdbcMigraiton` with the dependency injection mechanisms of Spring (`@Autowired`). Please have a look at the linked pull request and the tests how this is done.

If this pull request is accepted and integrated, then I would also add some documentation to the website project of Flyway on how to use this new resolver.

Thanks and regards,

Guy

kapil-malik:
Hi,
This is very useful for us as well. The PR looks pretty cool. Any updates on how soon can it be part of flyway builds? 
Till then we will override SpringJbdcMigrationResolver (like your code) as a workaround.

guylabs:
@kapil-malik Thanks and yes this would be really useful. 

@axelfontaine Thanks a lot for taking it up as a feature!

wimdeblauwe:
+1 for this!

crstalli:
@wimdeblauwe Thank you very much! That was extremely helpful!

marbot:
+1, that would be a very useful feature.

axelfontaine:
See also #869

geirgp:
+1

tpischke:
+1

paolodenti:
+1

K-Yosuke:
+1

zinefer:
When I use Spring JPA flyway migrations seem to run before the application context has been instantiated ... Does anyone know of a way around this, or how I can instantiate my repositories inside a flyway migration?

OldIMP:
We use a workaround from https://github.com/avehlies/spring-beans-flyway2 But I really don't recommend injecting data repo in migrations since the domain could have been changed.

zinefer:
Yes, I've tried that technique... But in my project, flyway migrations are being run before the application context - I get Null exceptions.

zinefer:
Infact, if you take that project (spring-beans-flyway2) and add `compile("org.springframework.boot:spring-boot-starter-data-jpa")` to the dependencies in `build.gradle` it will fail with a null pointer exception

OldIMP:
Sry I forgot to mention our configuration:
```java
@Configuration
[...]

PR #1063, opened 2015-07-22
Use Spring ApplicationContext to resolve SpringJdbcMigrations

guylabs:
Pull request for #1062.


reda-alaoui:
Hello,

Thank you for your PR.
I am currently trying to use it on my project.
But it seems that it doesn't manage spring proxies.

Indeed if we add ```@Transactional``` to our ```SpringJdbcMigration``` object, call to ```extractMigrationInfo``` fails.
I think it should pass the extracted bean proxied by Spring to ```extractMigrationInfo```.

Regards

guylabs:
Hi @reda-alaoui,

why do you add `@Transactional` to your `SpringJdbcMigration`s? Because these migrations are executed inside a transaction created by Flyway.

Thanks and regards,

Guy

wimdeblauwe:
I am trying to use this custom migrator in my own project. For this, I added `ApplicationContextAwareSpringJdbcMigrationResolver` in my own project until this will be released in Flyway itself. However, when using Flyway 4.0 (which I need to do to be able to disable the default MigrationResolvers), then the constructor is not valid. I needed to change it to:

    public ApplicationContextAwareSpringJdbcMigrationResolver(Scanner scanner, Location location, FlywayConfiguration configuration, ApplicationContext applicationContext) {
        super(scanner, location, configuration);
        this.applicationContext = applicationContext;
    }

nbijl:
Can we also use that to make use of repositories in the migrations?

wimdeblauwe:
I have used it with a Spring Data repository and it worked (for inserting
data). I did had to disable the hibernate ddl validation because it ran
before the flyway scripts.
Op vr 8 apr. 2016 om 09:38 schreef nbijl <notifications@github.com>

guylabs:
Hi @wimdeblauwe,

thanks for the hint. I just updated the pull request, merged the master to the branch and upgraded the Spring dependency to the latest version.

Regards,

Guy

alexparlett:
We found that there might be an issue with how this is done. If you keep the default migration resolvers and just use the application context aware one as a custom migration resolver then you cannot have the spring autowired and spring jdbc migrations in the classpath locations as the old spring jdbc migration resolver attempts to run the migrations as they have a base class of SpringJdbcMigration.

However if they arent on the flyway classpath and instead just use Springs autowiring to find them then you run into the issue that flyway is then unaware of all the migrations and may attempt to apply migrations in the wrong order. Ideally these shouldn't share the same base interface of SpringJdbcMigrations. We've made this change to our local version of flyway and are happy to provide a pr into either here or your fork Guy.

guylabs:
Hi @alexparlett 

yes that is true. Thanks for pointing this out! Well could you create a PR into my fork such that we have everything together for this feature? That would be awesome.

Thanks and regards,

Guy

geirgp:
What's the status on this PR? I want to inject Spring's Environment in order to have conditional flyway migrations for the purpose of enabling/disabling demo data.

geirgp:
@wimdeblauwe How did you disable the default MigrationResolvers? I'm using Spring Boot..

axelfontaine:

PR #1069, opened 2015-07-29
Decoupled DbSupportFactory so that it can be used as a hook to custom…

krishofmans:
Hi,

As we needed to customize the DbSupport for oracle and there was no real fix except extending the flyway class and overriding lots of methods to do so I decided to fix the root cause.

By moving this into a pluggable instance it can more easily be used to support newer database, even if the official flyway support is not yet ready. It would even be possible to start splitting out the db support so that the flyway-core can be less aware of all databases that it can support.

Kind regards,
Kris

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
@@ -30,6 +30,7 @@ import org.flywaydb.core.internal.command.DbSchemas;
 import org.flywaydb.core.internal.command.DbValidate;
 import org.flywaydb.core.internal.dbsupport.DbSupport;
 import org.flywaydb.core.internal.dbsupport.DbSupportFactory;
+import org.flywaydb.core.internal.dbsupport.DbSupportFactoryImpl;
 import org.flywaydb.core.internal.dbsupport.Schema;
 import org.flywaydb.core.internal.info.MigrationInfoServiceImpl;
 import org.flywaydb.core.internal.metadatatable.MetaDataTable;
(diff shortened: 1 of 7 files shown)


krishofmans:
Small example, I wanted to get the current logged in user of the application to be in the installed by field:

```java
        flyway.setDbSupportFactory(new DbSupportFactory() {
            @Override
            public DbSupport createDbSupport(final Connection connection, final boolean b) {
                return new OracleDbSupport(connection) {
                    @Override
                    public String getCurrentUserFunction() {
                        return "'" + FlywayService.this.sessionContext.getCallerPrincipal().getName() + "'";
                    }
                };
            }
        });
```

axelfontaine:
What was the reason you needed to customize things? Is there a potential improvement we should make in the main sources?

gertwillems:
Hey Axel,
we had a requirement to pass the user doing the flyway migrations (we wrote a little web application on top of flyway) into the database. So there was a need to set DbSupportFactory in our code,
kind regards,Gert
Date: Thu, 1 Oct 2015 10:58:34 -0700

frankhsueh:
hi hi ...  I might be able to add a bit of color.  at my workplace, we use statement-based replication => we can't use the USER() function that is "baked" into MySQLDBSupport.

https://dev.mysql.com/doc/refman/5.7/en/replication-rbr-safe-unsafe.html

pswrdf:
I really need this, because I want to change delimeter in DB2 sql sources, by set custom DELIMITER_KEYWORD, for example MySql style.
http://stackoverflow.com/questions/30897357/flyway-db2-change-delimiter-not-working

axelfontaine:
See #1151 which eliminates the vast majority of the reported needs for this. Closing.

Issue #655, opened 2013-12-03
Disable transaction for a particular migration


anthonygalea:
Since some SQL queries cannot run within a transaction (like adding an index concurrently in postgres) it would be useful to be able to turn transactions off when necessary.

The equivalent in ActiveRecord is disable_ddl_transaction! https://github.com/rails/rails/commit/b337390889cb4a9f80ed08daf072a043f0e7ddf3

falschparker82:
Any plans on integrating this into the SQL migrations API? If nobody is on it, we could have a go at it (or sponsor a bounty). Problem is of course, which interface should be used. Any preferences on the design side? Magic string inside the file name? SQL Comment inside the file?

axelfontaine:
@falschparker82 Ideally I would like to implement this so it is transparent to the user. Please contact me by email at "axel at boxfuse.com" so we can discuss your needs and the bounty details.

Cheers
Axel

yevah:
@axelfontaine thanks for nice feature. However we have a problem when having "create index concurrently...." queries in migrations. We are using postgresql and after this migration flyway just hangs. Actually index is created but in this point code execution just being kind of "paused" forever. Migration is not recorded in schema_version and running of next migration is not started.
It seems that problem appears because in this case Flyway has open transaction which locks metadata table. So maybe when executeIntransaction returns false, metadata table should not be locked too. You can easilty reporosuce the issue adding any "create index concurrently " statement to script file in your Junit test
Cheers,
Yeva

jimm-porch:
How do we tag a particular migration script so it gets run in auto-commit instead of in a transaction?  From the code change it looks like this can be disabled in Java but I would prefer if the migration script could control this.

For example the script could have a special name which triggers this, like ending in ".autocommit.sql" intead of ".sql", or the first line of the script could have a special comment like `-- @autocommit`, etc.

jimm-porch:
Another alternative would be to allow the `FlywayCallback` to control this, so that anyone could implement their own rules.

```java
/** @return true if this script should be run in a transaction, false to use auto-commit. */
default boolean isTransactional(Connection connection, URI script) = {
    return true;
}
```

jimm-porch:
[...]

PR #1071, opened 2015-08-02
Do not try to run migrations in transactions on databases that do not support transactional ddl

Nthalk:
Previously, in ticket #655, PostgresSQL was given special java api support to do things like `vacuum` that do not play nicely within transactions.

I currently have a script that renames a lot of constraints in Oracle, and apparently, there is some hidden limit to that within a transaction that causes the script to fault, yet run fine without a wrapping transaction.

However, #655 introduced `supportsDdlTransactions`, a good hint that perhaps, for migrating, we should not use a transaction, since an error will mean that we will most likely end up in an inconsistent state, and need other tools for rollbacks (I've been using vagrant && vagrant-snapshot).

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/command/DbMigrate.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/command/DbMigrate.java
@@ -278,7 +278,7 @@ public class DbMigrate {
             }
 
             final MigrationExecutor migrationExecutor = migration.getResolvedMigration().getExecutor();
-            if (migrationExecutor.executeInTransaction()) {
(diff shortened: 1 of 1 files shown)


axelfontaine:
Interesting idea. It is not without its downsides as migrations updating reference now also lose all rollbacks abilities.

Just to be sure: have you upgraded to the very latest Oracle version and verified it occurs there too?

Nthalk:
@axelfontaine Yes, but if the database doesn't `supportsDdlTransactions` there really is no reasonable assumption that migrations can even be rolled back (since most all of them can be assumed to modify structure).

Unfortunately, "Using the latest version of Oracle" is not a possibility. I've dived into their support forms, and done a bunch of research, in the end I just forked this project and am using that in my own maven repo.

Issue #1075, opened 2015-08-09
Support option to use Slf4j/Commons Logging/JUL when using the command line tool


ericparton:
Currently, when using the command line tool, Flyway only supports logging to standard out. It would be useful if the command line tool could optionally leverage Flyway's existing ability to use Slf4j or Commons Logging if their jars are on the classpath.

axelfontaine:
Thanks for your suggestion. What would be the real world use case for this?

Cheers
Axel

ericparton:
It would allow users of the command line tool to log the output of the tool to a file on disk if a jar with a logging configuration file and the logging jars were on the classpath. You could also do things like configure the logger to print the info statements to the console, but only log the debug statements to a file, so you don't have to run the tool with the debug flag and you still have the ability to consult the logs if there's an issue.

Thanks,
Eric

axelfontaine:
I ended up implementing it with an automatic switch instead.

Cheers
Axel

PR #1076, opened 2015-08-09
Support option to use Slf4j/Commons Logging/JUL when using the command line tool

ericparton:
Please see issue #1075


axelfontaine:
I ended up implementing it with an automatic switch instead.

Cheers
Axel

Issue #1078, opened 2015-08-11
Allow default migration resolvers to be overriden


martin-grofcik:
*Current behavior:*
* custom resolvers are added in CompositeMigrationResolver after the default one. (SqlMigrationResolver, JdbcMigrationResolver, SpringJdbcMigrationResolver) 
```java
            for (Location location : locations.getLocations()) {
                migrationResolvers.add(new SqlMigrationResolver(dbSupport, classLoader, location, placeholderReplacer,
                        encoding, sqlMigrationPrefix, sqlMigrationSeparator, sqlMigrationSuffix));
                migrationResolvers.add(new JdbcMigrationResolver(classLoader, location));

                if (new FeatureDetector(classLoader).isSpringJdbcAvailable()) {
                    migrationResolvers.add(new SpringJdbcMigrationResolver(classLoader, location));
                }
            }
            migrationResolvers.addAll(Arrays.asList(customMigrationResolvers));
```
*Problem description:*
When we want to override SpringJdbcMigrationResolver, we have no possibility to do that. Because SpringJdbcMigrationResolver is added to the resolvers by default and it is not possible to remove it. Spring JDBC migrations will be resolved twice (he first time from default resolver the second time from the default one) The usecase is not rare e.g.:
http://stackoverflow.com/questions/27105681/flyway-db-need-access-to-spring-environment-for-migration
https://github.com/flyway/flyway/pull/1062

*Solution proposals:*
* add default migration resolvers only in the case when there is no custom migration resolver
* another possibility is to do not add default migration resolver when there is already subclass of default migration resolver in the custom migration resolvers (little bit more complicated but allows backward comapatibility)

pauxus:
A third proposal: 

provide an additional flag parameter (flyway.skipDefaultResolvers) to allow to skip the default resolvers. This would also be completly backward compatible.

pauxus:
skipDefaultResolvers is actually not enough, since we need a couple things more:

- Own resolvers need to have access to Locations, Placeholders and DbSupport instances to fully take the place of the built-in resolvers.

I implemented this, along with the skipping in a PR: #811

pauxus:
created a new PR base on my previous patch:

#1183

pauxus:
Implementation of skipDefaultResolvers is merged with #1191. This allows to *replace* default resolvers, however custom subclasses of the resolvers are not yet supported.

axelfontaine:
Thanks @pauxus for the PR!

PR #1079, opened 2015-08-11
#1078 solution proposal 2 implementation

martin-grofcik:
proposed change allows to replace default MigrationResolvers. 
The implementation takes into account backward compatibilty


martin-grofcik:
inactive

Issue #735, opened 2014-04-24
cygwin classpath problem


lrozek:
In a cygwin shell under Windows Server 2008 R2 Standart try to run following: "./flyway" without quotes.

$ ./flyway
Error: Could not find or load main class org.flywaydb.commandline.Main

Related reference:
http://stackoverflow.com/questions/17433449/flyway-error-on-cygwin
https://github.com/flyway/flyway/issues/562
https://github.com/flyway/flyway/pull/603

PR #1124, opened 2015-11-06
Fix flyway script on Git Bash for Windows

sguillope:
New Git Bash for Windows (https://git-for-windows.github.io/) is recognized
as "MINGW..." system. For example on my test machine `uname` returns
the following:

```
$ uname
MINGW64_NT-6.1
```

This patch expands on the fix done in #735 to use Windows classpath
separator when running in Git Bash for Windows.

### DIFF
--- a/flyway-commandline/src/main/assembly/flyway
+++ b/flyway-commandline/src/main/assembly/flyway
@@ -19,7 +19,7 @@
 cygwin=false
 linux=false
 case "`uname`" in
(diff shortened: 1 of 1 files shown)


zaggino:
Hi guys, when would this be merged and released?

axelfontaine:
Thanks for reporting and fixing this! Truly appreciated! Merged.

Issue #1136, opened 2015-11-26
Multi Line comments are not handled correctly


dermojoo:
Flyway ( tried version 3.2.1 ) does not handle multiline comments inside ORACLE PLSQL Skripts correctly.
The first and second comment inside Package A is handled correctly the third is not. The third comment somehow escapes the package terminating / of package A.
After the migration this creates an invalid package A with the additional code from SOME_OTHER_PACKAGE_TO_FOLLOW inside.
Multi Line comments inside lines that start with a plain sql keywords (select, update, delete...) seam to work.
If the line starts with anything else it does not.

```
CREATE OR REPLACE 
PACKAGE BODY A
IS
PROCEDURE test
IS 
    test number;
BEGIN
    select count(*) into test /* working comment */ from dual;
	/* working comment */
    test:=/*not working comment*/1;
END;

END A;
/

CREATE OR REPLACE 
PACKAGE BODY SOME_OTHER_PACKAGE_TO_FOLLOW
....
...
..
.
```

dermojoo:
*update*
i made a mistake in my example code. The fist example ii posted did actually work. Now after my edit its not working and reproduces the error. The whitespaces before and after the opening/closing line comments made the code work. I debugged the code and drilled the issue down to org.flywaydb.core.internal.dbsupport.SqlStatementBuilder#applyStateChanges
the line 
TEST:=/*NOT WORKING COMMENT*/1;
is tokenized to
 0 = "TEST"
1 = ""
2 = "/*NOT"
3 = "WORKING"
4 = "COMMENT*/1"
and are parsed to this tokenTypes
0 = {SqlStatementBuilder$TokenType@1310} "OTHER"
1 = {SqlStatementBuilder$TokenType@1310} "OTHER"
2 = {SqlStatementBuilder$TokenType@1311} "MULTI_LINE_COMMENT"
3 = {SqlStatementBuilder$TokenType@1310} "OTHER"
4 = {SqlStatementBuilder$TokenType@1310} "OTHER"

So it missed the closing muli line comment.

axelfontaine:
Thank you very much for reporting this and providing a fix! Great work! Merged. (I've also added you to the hall of fame page (flywaydb 4.0 branch until merged))

PR #1140, opened 2015-12-02
fix for https://github.com/flyway/flyway/issues/1136

dermojoo:
and a small fix for a regex in OracleSqlStatementBuilder to avoid detection of Stings as sqlkeywords
and a improvement to skip opening multiline comments inside other multiline comments


javali:
top! needed!

sukisoft:
Really good piece of software here! we need this, too!

axelfontaine:
Thanks! Merged.

Issue #1154, opened 2016-01-07
Clean unable to drop function in db2


msallstr:
I'm getting the following db2 error when trying to run clean
```bash
[ERROR] Failed to execute goal org.flywaydb:flyway-maven-plugin:3.2.1:clean (default) on project uppfoljning: org.flywaydb.core.api.FlywayException: Unable to drop "TEST_FUNC(CHARACTER,CHARACTER): DB2 SQL Error: SQLCODE=-458, SQLSTATE=42883, SQLERRMC=UPPFOLJNING.TEST_FUNC, DRIVER=4.19.26 -> [Help 1]
```

From the [db2 documentation](https://www-01.ibm.com/support/knowledgecenter/SSEPEK_10.0.0/com.ibm.db2z10.doc.codes/src/tpc/n458.html)
_For example, assuming a function exists for which a parameter was defined as CHAR(12) on the CREATE FUNCTION statement, a reference to that function by a signature could specify this parameter as either CHAR(12), or CHAR(). The CHAR() syntax provides a way to say, "do not care about length, precision and scale attributes in finding a matching function"._

A simple test case
```sql
CREATE FUNCTION TEST_FUNC ( PARAM1 char(16), PARAM2 char(36) ) 
   RETURNS INTEGER 
   LANGUAGE SQL 
RETURN 1; 

drop function test_func(character, character); -- fails
drop function test_func(character(), character()); -- succeeds
drop function test_func(character(16), character(36)); -- succeeds
```

To never care about length may be the way to fix this.

Using flywaydb version 3.2.1 and DB2 v10.5.500.107 with fixpack 5

pswrdf:
DB2 v9.7 Flyway doesn't drops a function on clean operation (from java):

CREATE FUNCTION db2admin.get_version()
  RETURNS VARCHAR(16)
  SPECIFIC "GET_VERSION"
  NO EXTERNAL action F1:
  BEGIN atomic
    RETURN '7.0.0';
  END 

console show no errors:
1:44:06 PM org.flywaydb.core.internal.command.DbClean cleanSchema
INFO: Cleaned schema "DB2ADMIN" (execution time 00:00.016s).

But fails on followed by migrate operation: Found non-empty schema "DB2ADMIN" without metadata table!
And the function still exists in schema.

axelfontaine:
Fixed thanks to #1155 by @umefjord

PR #1155, opened 2016-01-08
#1154 - Drop functions with length attributes in DB2

umefjord:
FIX 1:
Patch to enable dropping functions in DB2 with datatypes that are not able to drop without parenthesis on the type, for example:

DROP FUNCTION MYFUNC(CHAR); --Fails => old implementation
DROP FUNCTION MYFUNC(CHAR()); --Succeeds => in this patch

Special case:

DROP FUNCTION MYFUNC(DECFLOAT); --Fails
DROP FUNCTION MYFUNC(DECFLOAT()); --Fails
DROP FUNCTION MYFUNC(DECFLOAT(16)); --Succeeds if length is 16 (can be 34) => not in this patch
DROP FUNCTION MYFUNC; --Succeeds, fallback workaround in this patch

FIX2:
Expression based indexes had to follow a special name convention in order to get flyway:clean to work.

CREATE TABLE CUSTOMER (
  firstname VARCHAR(25) NOT NULL
);

CREATE INDEX IDX_CUSTOMER_NAME ON CUSTOMER(LOWER(FIRSTNAME) ASC); --Success (due to special case)
CREATE INDEX IX_CUSTOMER_NAME ON CUSTOMER(LOWER(FIRSTNAME) ASC); --Fail to clean
CREATE INDEX I_CUSTOMER_NAME ON CUSTOMER(LOWER(FIRSTNAME) ASC); --Fail to clean

mvn flyway:clean

Gives the following error:

[ERROR] Failed to execute goal org.flywaydb:flyway-maven-plugin:3.2.1:clean (default-cli) on project db2flyway: org.flywaydb.core.api.FlywayException: Unable to clean
schema "DB2FLYWAY": DB2 SQL Error: SQLCODE=-658, SQLSTATE=42917, SQLERRMC=IX_CUSTOMER_NAME_V, DRIVER=4.19.26 -> [Help 1]

flyway:clean is trying to drop the materialized view that is created under the hood by the expression based index:

DROP VIEW IX_CUSTOMER_NAME_V; --Fail

...rather than dropping the index, i.e:

DROP INDEX IX_CUSTOMER_NAME; --Success

If the index name matches "IDX_%_V" it will succeed due to a special exclusion when views are fetched from SYSCAT. Not very robust and not compatible with other naming conventions. 

The fix widens the special exclusion to match "%_V" rather than "IDX_%_V".


axelfontaine:
Great work! Thanks! Merged.

PR #1162, opened 2016-01-13
#1078: skip default resolvers

pauxus:
This PR is a trivial implementation of #1078. It introduces a new parameter flyway.skipDefaultResolvers to allow working with custom resolvers, only.

It is base on my previous PR (#805) implementing a mechanism to provide the flyway configuration to callbacks and resolvers.


pauxus:
Closing since the relevant changes only make sense in combination with #805

PR #1183, opened 2016-01-26
Don't merge: #1078: Allow default resolvers to be replaced

pauxus:
This PR allows default resolvers to be completely replaced (#1078) (possibly by subclasses of the original classes). In order to do this, a couple of changes are necessary, which I will accumulate here:
- [x] Every information necessary for the resolvers should be injected instead of using the constructor. This allows custom resolvers and default resolvers to behave similarily.
- [x] Howevery, currently, three things can not be injected: `Location`, `DbSupport` and `Scanner` (because they are internal API/not available from `FlywayConfiguration`). I propose the following solutions for these:
  - [x] Change of the location handling of the default resolvers. Currently, one instance of each resolver is created per location. However, the current mechanism does not allow to inject a single location into a custom resolver. Thus, only one instance of each resolver (jdbc, spingJdbc, sql) should be created which should iterate the locations by itself. Since the locations as a whole CAN be injected via `FlywayConfiguration`, this would be sufficient
  - [x] change the `Scanner` caching mechanism. By not passing the Scanner around, but instead using a factory method with WeakReference based caching mechanism, the benefits of `Scanner` caching can still be reaped.
  - [x] ~~DbSupport is internal API. By providing a second interface (DbSupportAware), which itself is internal API, we could inject DbSupport into custom resolvers (replacements/subclasses of the default resolvers). This would, of course bear the risking of breaking DbAware custom resolvers with new versions of Flyway, which must be clearly stated.~~
  - [x] Remove usage of `DbSupport` in `SqlMigrationResolver` altogether. The Resolver and Executor do not really need it, they simple hand it down to SqlScript, which could always create a matching DbSupport object using the jdbcTemplate's connection during `execute()`. This would result in a lot more DbSupport objects being created, which could be (if necessary) negated by caching the `DbSupport` objects with their connection (`WeakReference in`WeakHashMap`). Also, now the parsing of the scripts would be done during the actual execution of the migration, not at resolving time (this could also be prevented by giving`SqlScript`access to the`FlywayConfiguration` object)
- [x] Built in custom resolvers must be skippable. This can either be done using a new property (`skipDefaultResolvers`), or by introducing three new properties defining the class names of the default resolvers. While the later would be somewhat more convenient for implementers, the first is IMHO cleaner.

The following two changes are part of a separate PR #1185
- [x] ~~Provide some methods of the default resolvers as protected, to allow overriding specific methods (Usecase for us: Parse the migration version not only from the filename, but also from the folder structure : 1.4_new_Feature/5_procedures/V1.2__demo_procedure --> 1.4.5.1.2)~~
- [x] ~~JdbcResolver and SpringJdbcResolver are almost identical. By extracting a common superclass, we could make it a lot easier to support similar frameworks.~~

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
@@ -1014,15 +1014,14 @@ public class Flyway implements FlywayConfiguration {
     /**
      * Creates the MigrationResolver.
      *
-     * @param dbSupport The database-specific support.
      * @return A new, fully configured, MigrationResolver instance.
      */
-    private MigrationResolver createMigrationResolver(DbSupport dbSupport) {
+    private MigrationResolver createMigrationResolver() {
         for (MigrationResolver resolver : resolvers) {
             ConfigurationInjectionUtils.injectFlywayConfiguration(resolver, this);
         }
 
-        return new CompositeMigrationResolver(dbSupport, this);
+        return new CompositeMigrationResolver(this);
     }
 
     /**
@@ -1235,7 +1234,7 @@ public class Flyway implements FlywayConfiguration {
 
             // force creation of a new Scanner to prevent location caching, because the classpath might have been changed
             Scanner scanner = Scanner.createNew(classLoader);
-            MigrationResolver migrationResolver = createMigrationResolver(dbSupport);
+            MigrationResolver migrationResolver = createMigrationResolver();
 
             if (callbacks.length == 0) {
                 setCallbacks(new SqlScriptFlywayCallback(dbSupport, scanner, locations, createPlaceholderReplacer(),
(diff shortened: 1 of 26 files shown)


pauxus:
Found a nasty classpath issue with the Scanner caching, which caused a single test case for Maven 3 to fail. Fixed that.

pauxus:
While taking a further look at the usage of DbSupport I think we can safely drop the DbSupportAware interface (and thus prevent mixing of internal and external API). The only place where DbSupport is really needed in the built in resolvers is in a PostgreSQL corner case in SqlScript. This corner case could (for now) be safely moved inside SqlScript.

axelfontaine:
No "kitchen-sink" PRs please. Try to keep it to exactly one change per PR. In the case of this one it should be ONLY the skipDefaultResolvers functionality. But then also implemented completely including the changes to the Ant, Maven, Gradle and SBT plugin as well as the Command-line tool including the sample config file. Also please post a matching PR against flywaydb.org for the docs.

Without this it is virtually impossible for me to merge this within simply investing less time by doing it all myself.

pauxus:
I will leave this one open as a place for discussion and create a new PR only for the skip part.

pauxus:
created PR #1191, website update follows

axelfontaine:
Closing old PR

PR #1196, opened 2016-02-02
PostgreSQL: trim schema name.

murdos:
The issue came while trying to detect schema in PostgreSQLDbSupport.getOriginalSchema() with an from originalSchema set with value: '"$user", public'.
I was ending with a schema named " public" (with a leading space).
The trim could either be done in PostgreSQLDbSupport.getOriginalSchema(), but probably twice, or just once in getSchema().


axelfontaine:
Thanks. I decided to fix it in PostgreSQLDbSupport.getOriginalSchema() instead as that was were the rest of the String manipulation was happening already.

murdos:
Fine, thanks.

PR #1198, opened 2016-02-03
Included abstract base classes for resolvers and migrations 

pauxus:
already implementing ConfigurationAware.

Changed javadoc to encourage usage of the abstract classes instead of implementing the interfaces directly.

Rationale: It makes sense to always implement ConfigurationAware, using a separate interface is basically to provide backward compatibility. Also, by using the abstract classes, we could transparently include new methods without breaking existing Migrations/Resolvers/Callbacks.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/api/callback/BaseFlywayCallback.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/api/callback/BaseFlywayCallback.java
@@ -16,14 +16,26 @@
 package org.flywaydb.core.api.callback;
 
 import org.flywaydb.core.api.MigrationInfo;
+import org.flywaydb.core.api.configuration.ConfigurationAware;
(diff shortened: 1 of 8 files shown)


axelfontaine:
Can you rename the classes to Base... instead of Abstract... for better consistency?

Also I would prefer a different wording for the interface javadoc changes. Think "for your convenience" instead of "not recommended"...

pauxus:
Renaming: Sure

The advantage of abstract classes instead of interfaces is backward compatibility. If a new method is introduced (in Flyway 5 for example), it would break all existing implementations. By encouraging implementers to subclass the base class instead, we can silently include default implementations for that methods to prevent that. Going that way would have made the ConfigurationAware part easier.

So, I would suggest leaving the wording as is and as of Flyway 4 start recommending to use the base classes instead.

Jenkins uses that technique with good success.

axelfontaine:
Thanks! Merged.

Issue #1200, opened 2016-02-05
Allow command line tool to set system properties


ericparton:
It would be useful for the command line tool to have the the ability to set system properties that could be used by custom Java migrations, callbacks, migration resolvers, etc.

axelfontaine:
Thanks for the suggestion. This can already easily be accomplished using environment variables instead. I therefore don't see the need to add this at this point.

PR #1201, opened 2016-02-05
Added ability to set Java system properties from the command line tool

ericparton:
Implementation for issue #1200. The command line tool can set system properties either through the use of the `-systemProperties.myproperty=myvalue` command line argument or by setting `flyway.systemProperties.myproperty=myvalue` in the configuration file.


axelfontaine:
Thanks for the pull request. See my comments on #1200

PR #1204, opened 2016-02-09
A test case for OSGI service-based migrations

renat-sabitov:
This is a test case proving that it is possible to create an
OSGI service to migrate completely separate bundle, using it's
classloader to locate resources and classes.


renat-sabitov:
@axelfontaine Can we have a new release please? I need this test to work in a released version.

renat-sabitov:
@axelfontaine is anything wrong with the test? I just want it to be merged to be sure that there's no regression

axelfontaine:
Closing as this only affects outdated code

Issue #738, opened 2014-04-25
Support multiple tasks of the same type in gradle


jnizet:
I have a unit-test database, and an integration-test database (and I could have more for production, for example), each with a different user/password/url. And I'd like to be able to create flyway tasks for each of them:

```
class PostgresDatabase {
    String name;
    String userName;
    String password;
    String url;

    PostgresDatabase(String name, String userName, String password, String url) {
        this.name = name;
        this.userName = userName
        this.password = password;
        this.url = url;
    }
}

def unit = new PostgresDatabase('unit', 'foo_unit', 'foo_unit', 'jdbc:postgresql://localhost/foo_unit');
def integ = new PostgresDatabase('integ', 'foo_integ', 'foo_integ', 'jdbc:postgresql://localhost/foo_integ');

[integDatabase, unitDatabase].each {db ->
    project.task('flywayMigrate' + db.name.capitalize(), type: org.flywaydb.gradle.task.FlywayMigrateTask) {
        ...
    }
}
```

Unfortunately, that's not as easy as it could be, because the Flyway gradle tasks don't have any property. They all read their properties from a unique flyway extension.

It would be nice to be able to create tasks having unique properties, overriding those from the extension.

It would also be nice to be able to use a flyway-base-plugin, which would only make the tasks available (for my use-case, for example), and a flyway-plugin, relying on the base plugin, and which would add default tasks using the extension.

I've managed to do it by subclassing the Flyway tasks and overriding the `createFlyway()` method, but that's painful, and relies on an undocumented implementation detail.

netmikey:
I don't think there's a need for 2 separate plugins. As Peter Niederwieser (a Gradle core dev) points out in this SO question http://stackoverflow.com/a/18142219/737587 

> Basically the task class(es) need to declare all properties (or more) that the extension has, using the extension properties as defaults. Have a look at the code quality plugins in the Gradle codebase (e.g. `CheckstylePlugin` or `FindbugsPlugin`). By convention, concrete task classes don't have `Task` in their name.

Implementing this would solve all your problems: it would make the current syntax continue working without any change, and it would allow ppl to create new flyway tasks and configure them ad-hoc, effectively overriding the defaults from the extension.

ben-manes:
If this is reintroduced, the [pull request](https://github.com/ben-manes/gradle-flyway-plugin/pull/2) where this feature was originally implemented may be a good reference for whomever resolves this issue.

alkamo:
[...]

PR #1208, opened 2016-02-15
Support for multiple databases/targets in Gradle

alkamo:
This is a stepping stone to resolving #738. While the feature requested is not yet implemented, the ability to support multiple configurations in a single run will enable us to eventually run tasks for the individual configurations separately.

Care was taken in this implementation to not change the existing behavior of the Gradle plugin. The simplest case still works.

If multiple databases are configured, then the "master" properties become defaults that are used only if the individual databases do not have a value assigned to the property.

Additionally, there are now "instruction" properties that correspond to each of the array/map properties. These allow the user to change the default "prioritization" behavior to a "merge" behavior. When set to "merge" properties set in different levels are combined, rather that taken in the order they are found. For instance, with the placeholders Instruction set to "merge", if you specify "flyway.placeholders.stub_a=a" as a environment variable and "placeholders['stub_b':b']" in the build.gradle file, both placeholders will be resolved (as opposed to the current implementation, where only "stub_a" would be resolved).

### DIFF
--- a/flyway-gradle-plugin-largetest/src/test/java/org/flywaydb/gradle/largetest/GradleLargeTest.java
+++ b/flyway-gradle-plugin-largetest/src/test/java/org/flywaydb/gradle/largetest/GradleLargeTest.java
@@ -49,7 +49,7 @@ public class GradleLargeTest {
     @Test
     public void error() throws Exception {
         String stdOut = runGradle(0, "error", "clean", "flywayMigrate");
-        assertTrue(stdOut.contains("Successfully validated 0 migrations"));
+        assertTrue(stdOut.contains("No migration necessary."));
     }
 
     /**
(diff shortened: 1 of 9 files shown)


alkamo:
The next iteration of this is complete, but has not been pushed because it adds a significant number of files and I wanted to limit the scope of this pull request.

I believe that code will resolve #738.

alkamo:
I'm working on updating this PR, given that the Groovy has been excised.

@axelfontaine I'm wondering about the attributes that were copied from FlywayExtension to AbstractFlywayTask. That amount of copy/paste always bothers me. Is there a reason that you didn't just add a FlywayExtension object with AbstractFlywayTask?

alkamo:
This PR has been updated to be Java-only.

Additional changes:

Attributes that mirror FlywayExtensionBase in AbstractTask have been removed and FlywayExtensionBase attribute has been added. I'm not sure what the purpose of this is, but this seems like the cleaner implementation.

The createFlyway method and methods that support it have been moved out of AbstractFlywayTask and added to a new class, FlywayCreator. They didn't really make sense in AbstractFlywayTask in the first place and creating a dedicated factory will make unit testing easier. This also makes AbstractFlywayTask more concise.

As in the earlier PR, the ability to choose between merging and overriding lists/maps at different levels has been added.

axelfontaine:
This has been resolved in a different way a while ago in Flyway 4.1.0.

Issue #1117, opened 2015-10-28
Transparently upgrade metadata table from 3.x to 4.x format


axelfontaine:
- [x] MigrationType INIT -> BASELINE
- [x] Recalculate checksums by ignoring line endings
- [x] Drop version_rank column
- [ ] Make version nullable
- [ ] Make installed_rank the new primary key
- [ ] Drop installed_rank index
- [ ] Drop version_rank index

numero-six:
> Recalculate checksums by ignoring line endings

In my opinion, we should also ignore comments. Many copyright and licensing policies require to place a license header on every file of a project. It means that if the project using Flyway has its license changed, all checksum will changes, raising Flyway errors.

axelfontaine:
@cteig Flyway 4.0 will be released next week. Can you check **this week** that `MigrationTestCase.upgradeMetadataTableTo40Format()` passes for DB2 zOS to ensure those users will have a smooth migration path to Flyway 4.0.

axelfontaine:
@nathanvick Flyway 4.0 will be released next week. Can you check **this week** that `MigrationTestCase.upgradeMetadataTableTo40Format()` passes for Redshift & Vertica to ensure those users will have a smooth migration path to Flyway 4.0.

axelfontaine:
@MichaelF25 Flyway 4.0 will be released next week. Can you check **this week** that `MigrationTestCase.upgradeMetadataTableTo40Format()` passes for SolidDB to ensure those users will have a smooth migration path to Flyway 4.0.

nathanvick:
@axelfontaine,

I regret that I've been feeling sick this week. I hope get to it early this
week.

-Nathan

axelfontaine:
@MichaelF25 Did you have a chance to check this?

axelfontaine:
@nathanvick No worries. Worst case this will be for 4.0.1

MichaelF25:
@axelfontaine I am now working for a different company and we do not use SolidDB. I've handed over this issue to my former colleagues and hopefully one of them will provide an update about this issue soon.

PR #1210, opened 2016-02-16
Updated table space for meta table, with db2 z/OS standard values

cteig:
Hi, 

The pull request contains changes in meta data table. Encoding is changed to EBCDIC, the standard for zOS, and the tablespace are updated with more standard values. 

Hope to reach the 4.0 release!

Christine


axelfontaine:
Hi Christine,

Can you provide a smooth migration path for existing users (#1117) that migrates the encoding for them?

Best regards,
Axel

cteig:
Hi

I have looked at updateMetaDataTable.sql for db2zos. I don´t think "call ADMIN_CMD" is something we can use. I have changed job and have no test environment anymore.  But I know someone who can test and change the new version, but we need a day or two. We will also look at a migration path, but I´m not sure if we need it. Maybe I will remove the encoding from the create table statement, and instead use the encoding scheme of the database. 

Christine

axelfontaine:
Yes, I think simply using the encoding of the DB is probably the best option.

I simply copied the regular DB2 update file, so no worries about call ADMIN_CMD. All I care about is that the test passes :-)

axelfontaine:
@cteig Did you make progress on this?

cteig:
I have an appointment with someone with a test-environment tomorrow. 
The upgrade was not so easy as we first assumed. It has to work with the db2 z/os version 9 and newer
It is not possible to delete columns in version 9 :-(

Christine

axelfontaine:
No worries Christine, if it doesn't work we'll simply instruct users to drop the table and recreate it with Flyway.baseline()

cteig:
Hi

The issue is now handed over to a former colleague. You will get a pull request for this change, but I can´t say when. Mayby for a later release.

Christine

axelfontaine:
Obsolete. Closing.

Issue #1211, opened 2016-02-17
PostgreSQL upper case schema names end up double quoted


smohara:
If the first element in the search_path is a quoted upper case string, the doGetCurrentSchemaName method will return this quoted string as the schema name. When the schema name is then used in the creation of the schema_version table, the resulting sql is 
CREATE TABLE ""MYSCHEMA""."schema_version" (
    "installed_rank" INT NOT NULL,
    "version" VARCHAR(50),
    "description" VARCHAR(200) NOT NULL,
    "type" VARCHAR(20) NOT NULL,
    "script" VARCHAR(1000) NOT NULL,
    "checksum" INTEGER,
    "installed_by" VARCHAR(100) NOT NULL,
    "installed_on" TIMESTAMP NOT NULL DEFAULT now(),
    "execution_time" INTEGER NOT NULL,
    "success" BOOLEAN NOT NULL
) WITH (
  OIDS=FALSE
)

PR #1212, opened 2016-02-17
#1211 Postgres upper case schema fix

smohara:
Remove the quotes from a schema name returned by the doGetCurrentSchemaName  in PostgreSQLDbSupport.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/dbsupport/postgresql/PostgreSQLDbSupport.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/dbsupport/postgresql/PostgreSQLDbSupport.java
@@ -1,12 +1,12 @@
 /**
  * Copyright 2010-2016 Boxfuse GmbH
- *
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
- *         http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


flyway-core/src/test/java/org/flywaydb/core/internal/dbsupport/postgresql/PostgreSQLDbSupportSmallTest.java
```diff
@@ -26,4 +26,14 @@ public void doQuote() {
         assertEquals("\"abc\"", dbSupport.doQuote("abc"));
         assertEquals("\"a\"\"b\"\"c\"", dbSupport.doQuote("a\"b\"c"));
     }
+
+    @Test
+    public void undoQuote() {
+        PostgreSQLDbSupport dbSupport = new PostgreSQLDbSupport(null);
+        assertEquals("abc", dbSupport.unQuote("\"abc\""));
+        assertEquals("abc", dbSupport.unQuote("abc\""));
```
axelfontaine: Why is this and all other cases from line 34 to 37 necessary?

flyway-core/src/main/java/org/flywaydb/core/internal/dbsupport/postgresql/PostgreSQLDbSupport.java
```diff
@@ -119,6 +119,14 @@ public String doQuote(String identifier) {
         return "\"" + StringUtils.replaceAll(identifier, "\"", "\"\"") + "\"";
     }
 
+    public String unQuote(final String quoted) {
```
axelfontaine: Can you make this package-private and add JavaDoc?

axelfontaine:
I am having doubts about this fix. How does it deal with multiple schemas in the search path?

smohara:
It doesn't work well when multiple schema are returned. 
The method this is used in is called doGetCurrentSchemaName, which suggests it will return a single schema name. Prior to my change it would have returned multiple schemas. I assume we want to treat the first element in the search path as the current schema?

axelfontaine:
It is actually there to store the current schema of a connection and restore it afterwards. For PG this actually means storing the search_path and restoring it. The name is misleading for PG but correct for the other DBs.

axelfontaine:
Closing due to lack of response from original reporter

PR #1217, opened 2016-02-26
Fixed flyway 4.0 upgrade for Redshift (it does not support alter column)

nathanvick:
@axelfontaine 

I fixed the flyway 4 metadata upgrade for Redshift. Redshift does not support ALTER COLUMN, so we have to create new columns and copy the data instead.


axelfontaine:
Thanks!

PR #1237, opened 2016-03-10
Fixed #1120, 3.x->4.0 migration, storage group setting, moved tests

### LINKED ISSUE #1120: avoid 'create tablespace' for DB2 on zOS, allow an encoding parameter
Hallo

we've tweaked around to get flyway running on our DB2 on zOS environment. We have noticed some things that might be bugs (but frankly: I'm not sure).

The migrate goal tries to create the versions table and a tablespace: see [createMetaDataTable.sql](https://github.com/flyway/flyway/blob/master/flyway-core/src/main/resources/org/flywaydb/core/internal/dbsupport/db2zos/createMetaDataTable.sql).

we have removed the create tablespace statement (since the other scripts do not contain that either) and changed the create table statement as follows (our complete createMetaDataTable.sql):

```sql
set current sqlid = '${schema}';


CREATE TABLE "${schema}"."${table}" (
    "version_rank" INT NOT NULL,
    "installed_rank" INT NOT NULL,
    "version" VARCHAR(50) NOT NULL,
    "description" VARCHAR(200) NOT NULL,
    "type" VARCHAR(20) NOT NULL,
    "script" VARCHAR(1000) NOT NULL,
    "checksum" INT,
    "installed_by" VARCHAR(100) NOT NULL,
    "installed_on" TIMESTAMP NOT NULL WITH DEFAULT,
    "execution_time" INT NOT NULL,
    "success" SMALLINT NOT NULL,
    CONSTRAINT "${table}_s" CHECK ("success" in(0,1))
)
IN database "${schema}"
  CCSID EBCDIC
;

-- Indekser og constraints
CREATE UNIQUE INDEX "${schema}"."${table}_VPK_IDX" ON ${schema}."${table}" ("version" ASC);
ALTER TABLE "${schema}"."${table}" ADD CONSTRAINT "${table}_PK" PRIMARY KEY ("version");

CREATE INDEX "${schema}"."${table}_VR_IDX" ON "${schema}"."${table}" ("version_rank");
CREATE INDEX "${schema}"."${table}_IR_IDX" ON "${schema}"."${table}" ("installed_rank");
CREATE INDEX "${schema}"."${table}_S_IDX" ON "${schema}"."${table}" ("success");
```

Notice: we use CCSID EBCDIC instead of CCSID UNICODE (flyway version 3.2.1). That might needs to be a placeholder as well (for our database it's mandatory to use EBCDIC). And there is a IN database at the end.

The DB2 zOS Version and driver we use is:
```
DB2
DSN11015
IBM Data Server Driver for JDBC and SQLJ
4.17.29
```

Thanks a lot!

Muni10:
Fixed #1120: Removed CCSID to allow DB2 to use default encoding
Fixed upgrade from 3.x to 4.0 Flyway schema_version for DB2 z/OS
Removed STGROUP definition in CREATE TABLESPACE to allow database to use schema default.
Moved tests to DB2zOSMigrationMediumTest overriding tests in MigrationTestCase to allow tests to run without having DB2-user with excessive permissions.


flyway-core/src/test/java/org/flywaydb/core/internal/resolver/sql/SqlMigrationResolverSmallTest.java
```diff
@@ -15,6 +15,14 @@
  */
```
axelfontaine: Please revert these unnecessary formatting changes. They only make the PR harder to review. Thanks!

axelfontaine:
Thanks! Merged.

PR #1242, opened 2016-03-15
Enhancement for #724: Prompt for user/password if property exists but is empty (for default .conf setup).

PredatorVI:
The default flyway.conf configuration has the following defaults:

```
# User to use to connect to the database (default: <<null>>)
flyway.user=

# Password to use to connect to the database (default: <<null>>)
flyway.password=
```

The code would not prompt for username/password in this case because the original check was only looking for flyway.user or flyway.password in the property map, but didn't account for empty values.


axelfontaine:
No, it doesn't. See https://github.com/flyway/flyway/blob/master/flyway-commandline/src/main/assembly/flyway.conf#L46

This was changed in 4.0 for exactly this reason.

Cheers
Axel

PR #1246, opened 2016-03-18
Fixed bug in properties configuration. 

caiodangelo:
The property flyway.placeholderReplacement wasn't being handled correctly in flyway-core.

If we use the maven plugin the property is set and the behaviour is the expected. Otherwise, don't.


axelfontaine:
Duplicate of #1233

PR #1247, opened 2016-03-19
add an autoUpdateMetaDataTable property

cheister:
Add an autoUpdateMetaDataTable property that determines whether Flyway should automatically update the schema_version table to the latest format.   

We need this for some of our databases that get updated offline but still use Flyway to determine whether migrations are up-to-date.  I don't expect many of the Flyway methods to work with the old schema_version table but it will be useful to have Flyway fail and not have attempted to modify the schema_version table to the new version.


axelfontaine:
Thank you for your PR, but this is not something I want to pursue at this point. If you want to determine if migrations are up to date, you can still use the old version of Flyway to do so.

Issue #1231, opened 2016-03-04
upgrading AWS Redshift DB from Flyway 3.2.1 to 4.0 failed - couldn't upgrade metadata table


wgillett:
###### What version of Flyway are you using?

3.2.1

###### What database are you using (type & version)?

AWS Redshift

###### What operating system are you using?

Linux

###### What did you do?
(Please include the content causing the issue, any relevant configuration settings, and the command you ran)

Changed my code to use Flyway 4.0. On calling Flyway#migrate, Flyway tried to upgrade and failed (details below).

###### What did you expect to see?

Expected to see successful Flyway init.

###### What did you see instead?

Flyway init failed:

2016-03-03 23:09:14,251 INFO  [main] o.f.c.i.m.MetaDataTableImpl - Upgrading metadata table "public"."schema_version" to the Flyway 4.0 format
...
org.flywaydb.core.internal.dbsupport.FlywaySqlScriptException: 
Script failed
-------------
SQL State  : XX000
Error Code : 500310
Message    : [Amazon](500310) Invalid operation: cannot insert/update into table after dropping non-nullable column;
Line       : 20
Statement  : UPDATE "public"."schema_version" SET "version_temp"="version"
	at org.flywaydb.core.internal.dbsupport.SqlScript.execute(SqlScript.java:117) 
	at org.flywaydb.core.internal.metadatatable.MetaDataTableImpl.upgradeIfNecessary(MetaDataTableImpl.java:79) 
	at org.flywaydb.core.Flyway.execute(Flyway.java:1356) 
	at org.flywaydb.core.Flyway.migrate(Flyway.java:917)

axelfontaine:
@nathanvick Can you have a look at this? Thanks!

nathanvick:
Can you tell me what version of Redshift is being used? The error is not familiar to me, but it makes me wonder if it is an old version of Redshift. I have been dropping not null columns and then updating data for a couple years.

http://docs.aws.amazon.com/redshift/latest/dg/r_VERSION.html

PS: I tested this with flyway 4 (command line), upgrading from flyway 3.2 (command line), and it worked for me,  using both the Postgres jdbc driver and the Redshift jdbc driver.

wgillett:
Hmmm. Not aware there was a choice of Redshift version, it's SaaS. See below for version info nonetheless. I'm invoking Flyway through the API rather than the command line. 

[...]

PR #1256, opened 2016-03-24
Fixed issue with upgrading metadata table in AWS Redshift #1231

nathanvick:
This fixes the issue where upgrading the metadata table (from Flyway v3.2.1 to v4.0) failed, for at least some installations of AWS Redshift. Redshift does not support altering columns within a table. There are two common workarounds for this:

* **Copying data between temp columns:** adding a temp column, copying data to the temp column, dropping the original column, adding the updated column, copying data back to the updated column, dropping the temp column
* **Copying data between temp tables:** renaming the existing table, creating the updated table, copying data to the updated table, dropping the renamed table

This pull request changes the approach from "copying data between temp columns" to "copying data between temp tables" because the former approach requires a commit half way through the process, for some installations of Redshift.


axelfontaine:
Thanks Nathan! Merged.

axelfontaine:
@nathanvick Could you please contact me by email at axel at boxfuse dot com ? Thanks!

Issue #1218, opened 2016-02-29
Command line interface does not work without "drivers" folder present


biran0079:
I am trying to invoke flyway command line interface through `flyway-commandline.jar`, but it complains about not finding `drivers` folder. 
I did pass jdbc driver class through class path. I think this should be a legit use case, and the `drivers` folder should be optional.

axelfontaine:
Thanks for reporting this. What was the reason you needed another driver? And how exactly did this fail?

ricardo-s-ferreira:
I've the same problem. For organization issue I want put all jar files in one folder called lib, but the command line failed if the folder drivers doesn't exist. "Directory for Jdbc Drivers not found: ..../drivers"

axelfontaine:
You can leave an empty folder to save space or use the Java API if you want more control.

Won't fix.

PR #1257, opened 2016-03-28
Command line interface does not work without "drivers" folder present #1218 

ricardo-s-ferreira:
Allow the command line interface to work without the "drivers" folder. the drivers are search in path "flyway.jarDirs" and in classpath.


axelfontaine:
This introduces a potential NPE for the Java Migrations directory. Also there is really no valid reason the delete the drivers directory (just keep an empty one if you want to save space).

Won't merge.

Issue #1259, opened 2016-03-29
Support for SAP (Sybase) SQL Anywhere


ssindelar:
SAP (Sybase) SQL Anywhere is the smaller less popular cousin of SAP (Sybase) Adaptive Server which is supported since Flyway 4.0. The databases are pretty similar (structure and syntax). 

Due to legacy issues we still need to support SQL Anywhere. To make it less a hassle I added support for SQL Anywhere to Flyway and would like to share my work.

See pull request for implementation details.

PS: I hope it is ok to ignore the template, since it seems to be for bug reports.

rtoengi:
One of our projects has to support SAP SQL Anywhere 16, so support for it is highly appreciated.

ssindelar:
If you like you can use my 4.0 branch with SQL Anywhere support. We are using it in Production for a while and didn't had any issues with Anywhere 12.

https://github.com/ssindelar/flyway

PR #1260, opened 2016-03-29
#1259: Support for SAP (Sybase) SQL Anywhere

ssindelar:
Pull Request to address issue #1259 .

The implementation is based on the SAP (Sybase) ASE implentation. The databases are similar in structure and syntax. I needed to make minor modifications to the ASE Database Support to make it reuseable. The major difference is that when you use SQL Anywhere as Database admin (dba account) the Flyway clean also droped essential system/metadata tables. Therefore, only tables that are created by the current user are considered, when analysing the database, which excludes the system/metadata tables

I could just tested it on SQL Anywhere 12.0. The database is only slowly developed and I don't expect any problems with newer versions.

The implementation can be tested using the [SAP SQL Anywhere Developer Edition](http://scn.sap.com/community/sql-anywhere). An empty database (flyway_test) needs to be created. The Developer Edition should also include the jdbc driver.

There are currently two problems: 
- SQL Anywhere can be used with an odbc-based driver or a preferred pure jdbc driver (jconnect). The problem is the jconnect driver is not public and I couldn't figure out the correct group/artifactId. We are using an internal artifact. So I didn't add a dependency to the driver to the pom.
- The contribution guidelines mention the documentation needs to be updated, but what exactly needs to be done.

This is my first open soure contribution. I hope I didn't forgot anything essential.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/dbsupport/DbSupportFactory.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/dbsupport/DbSupportFactory.java
@@ -32,6 +32,7 @@ import org.flywaydb.core.internal.dbsupport.saphana.SapHanaDbSupport;
 import org.flywaydb.core.internal.dbsupport.solid.SolidDbSupport;
 import org.flywaydb.core.internal.dbsupport.sqlite.SQLiteDbSupport;
 import org.flywaydb.core.internal.dbsupport.sqlserver.SQLServerDbSupport;
+import org.flywaydb.core.internal.dbsupport.sybase.asa.SybaseASADbSupport;
 import org.flywaydb.core.internal.dbsupport.sybase.ase.SybaseASEDbSupport;
 import org.flywaydb.core.internal.dbsupport.vertica.VerticaDbSupport;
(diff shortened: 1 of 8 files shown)


ssindelar:
We the changes in production for a couple of months now and no problems yet.

axelfontaine:
This needs some cleanup before it can be merged.

Some things to consider:

Can you add the concurrent test and make sure it passes?
Can you ensure the tests only run with a dedicated Maven profile?
Can you also add the url autodetection to the datasource as well as url examples to the flyway.conf file?
Do you agree to maintain and support this in the future?
Can you also submit a PR with the relevant changes against flyway/flywaydb.org? (dedicated db page, adjusted navigation in menu and previous/next pages, listed on front page, setup instructions)

Thanks!

ssindelar:
In the mean time we decided to drop the SQL Anywhere Support from our Product. This puts me in the position that I can no longer support this PR in the future during work. I would have to do it in my free time for a database I dislike very much. 

I am sorry, but I think it is better to reject the pull request and hope that no other poor soul needs to use SQL Anywhere.

axelfontaine:
Fair enough. Closing for now until a new maintainer steps up.

Issue #1265, opened 2016-04-01
Upgrade metadata table to v4.0 fails on Derby


azalesky:
###### What version of Flyway are you using?
4.0

###### What database are you using (type & version)?
Apache Derby 10.10.2.0 - (1582446)

###### What operating system are you using?
CentOS7

###### What did you do?
I try to switch from v3.2.1 to v4.0

###### What did you expect to see?
Metadata table should be altered on first migrate call

###### What did you see instead?
Error code 30000, SQL state 42X01: Syntax error: Encountered "NOT" at line 1, column 69.

axelfontaine:
Thanks for the fix! Merged.

PR #1266, opened 2016-04-01
#1265 fix upgrade metadata table for Derby

azalesky:
The problem is that Derby has no command ALTER COLUMN column-name DROP NOT NULL
But it has ALTER COLUMN column-name NOT NULL


axelfontaine:
This actually does the opposite of what is needed. The not null constraint should be dropped not enabled.

PR #1282, opened 2016-04-15
Add support for PostgresPlus Advanced Server

rbellamy:
[EnterpriseDB Postgres Plus Advanced Server](http://www.enterprisedb.com/products-services-training/products/postgres-plus-advanced-server)

* New dialect supporting the weird blend of Oracle and PostgreSQL known as SPL.
* Unit tests

Unit tests pass

75 tests pass, 4 ignored:
* xml types and indexes are not tested
* text types and indexes are not tested
* pg_copy JDBC functionality is not yet available
* lock - looks like maybe there is a overly restrictive LOCK MODE by default in EDB

There were other tests that had to be commented out because the EDB JDBC drivers have some
additional bugs:
* TYPE BODY cannot be executed.
* Package Body with complex comments causes ArrayOutOfBoundsException.

And one test that just plain confused me:
* The Oracle tests I used as my starting point included a V2__Invalid.sql migration script which had an
invalid procedure in it - couldn't figure out why, so it's not being tested.

### DIFF
--- a/flyway-core/src/test/resources/migration/dbsupport/postgresplus/sql/package/~V3__ComplexComments.sql
+++ b/flyway-core/src/test/resources/migration/dbsupport/postgresplus/sql/package/~V3__ComplexComments.sql
@@ -1,0 +1,95 @@
+--
+-- Copyright 2010-2016 Boxfuse GmbH
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--         http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+
+create or replace
+PACKAGE AgingUtils
+IS
+  /* ----------------------------------
+  NAME
+  AgingUtils - create / drop partitions.
+
+  DESCRIPTION
+
+  NOTES
+  load_defaults - procedure with default ranges.
+  Below prefixTsName. All tablespaces will have name started with this prefix.
+
+  Important:
+  Fixes for NETRO schema.
+	Run output from those queries.
+SQL> select 'alter table '||table_name||' rename partition '||partition_name||' to '||replace(partition_name,'GG','G')||';' from user_tab_partitions where partition_name like '%GG%';
+SQL> select 'alter index '||index_name||' rename partition '||partition_name||' to '||replace(partition_name,'GG','G')||';' from user_ind_partitions where partition_name like '%GG%';
+----------------------------------  */
+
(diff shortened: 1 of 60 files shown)


flyway-core/src/test/resources/migration/dbsupport/postgresplus/sql/package/~V3__ComplexComments.sql
```diff
@@ -0,0 +1,95 @@
+--
```
axelfontaine: Why does this filename start with ~ ?

flyway-core/src/test/resources/migration/dbsupport/postgresplus/sql/procedure/~V2__Invalid.sql
```diff
@@ -0,0 +1,21 @@
+--
```
axelfontaine: Same here: why the ~ in the file name?

axelfontaine:
Thanks for contributing this.

A few points are still open:

- What are those files starting with a ~ ?
- Shouldn't this dialect be called EnterpriseDB or EDB since that is how everyone seems to refer to this offering? In the official marketing materials the only reference to Postgres Plus seems to be the cloud offering. Could you rename accordingly?
- Could you also contribute a documentation PR against the flyway-4.1 branch of flyway/flywaydb.org with updates to the homepage, DB setup page, docs, getting started & about overview pages as well as a dedicated docs page for the DB (with the necessary changes to the navigation)?
- Would you be willing to support this in the future should issues come up? (I'll do the triaging and will ping you if necessary)

Thanks!

flyway-core/src/test/resources/migration/dbsupport/postgresplus/sql/procedure/~V2__Invalid.sql
rbellamy: As I mentioned in my initial comments, I just could not figure out why the Oracle tests included this "V2_Invalid.sql" test - it's essentially designed to fail, which fails the whole procedure test. I decided, against convention, to keep the file in the test scripts, rather than just deleting it - I was hoping someone looking at this PR could tell me one way or another why I might need it...

flyway-core/src/test/resources/migration/dbsupport/postgresplus/sql/package/~V3__ComplexComments.sql
rbellamy: This file is "disabled" because it triggers a bug in the EDB JDBC drivers - EDB knows about it, and I expect will fix the bug sometime in the near-to-somewhat-near future. The test itself is valid, once the bug is fixed.

rbellamy:
@axelfontaine to your points:

* I've answered the specific questions about the "disabled" test scripts - those are easily removed if that's your preference.
* I'll work on the docs and send a PR shortly.
* Yes, I will this code.

rbellamy:
And yes, EDB is a better name. Once we're done with this conversation here, go ahead and reject this PR and I'll rename and re-issue a new PR.

rbellamy:
I've got a new branch with the rename, ready when you are... I'm also working on the docs.

axelfontaine:
Sounds good please remove the disabled scripts and resubmit. Many thanks for this!

Issue #979, opened 2015-03-29
can't migrate as long as there are compile errors


axaluss:
I try to flywayMigrate but there are compile errors. 
So the migrate task is never executed.

The funny thing is, that i can only generate the proper code with jooq **after** i migrated the database.

Is there a way get around the compile dependency?

axaluss:
have a look at this fix: https://github.com/axaluss/FlywaySbt

axelfontaine:
Thanks for the report. This is something I plan to change for all build tool clients (ant, maven, gradle, sbt) in time for 4.0

nkgm:
+1
On maven a workaround is possible. SBT however requires more manual labor.

axelfontaine:
The default location for Maven, Gradle and SBT has been change to filesystem:src/main/resources/db/migration which should improve integration with jOOQ and remove the need for compilation when only using SQL files

nafg:
The fix breaks sbt subprojects (I had to add `flywayLocations := Seq("db/migration")`) since they are in a different subdirectory. I guess it would also not work with Play's directory layout or any other customization.

mikebridge:
Is there an example of how to set this up in sbt?  I still have a dependency on compilation in a subproject with 4.0.3:

```scala
lazy val `my-subproject` = (project in file("my-subproject"))
 .settings(
    flywayUrl := "jdbc:postgresql://localhost:5432/mydb",
    flywayLocations := Seq("db/migration"),
    flywayUser := "myuser")
```

nafg:
@mikebridge what does `inspect tree flywayMigrate` say the dependency is?

Also why do you need to override `flywayLocations`?

mikebridge:
@nafg It looks like flywayMigrate depends on fullClasspath.

As for `flywayLocations`,  I believe I applied your May 30 workaround for subprojects

```
> inspect tree identity-manager/flywayMigrate
[info] identity-manager/*:flywayMigrate = Task[Unit]
[info]   +-identity-manager/*:flywayConfig = Task[org.flywaydb.sbt.FlywayPlugin$Config]
[info]   +-identity-manager/runtime:fullClasspath = Task[scala.collection.Seq[sbt.Attributed[java.io.File]]]
[info]     +-identity-manager/runtime:exportedProducts = Task[scala.collection.Seq[sbt.Attributed[java.io.Fil..
[info]     | +-identity-manager/compile:packageBin::artifact = Artifact(identity-manager,jar,jar,None,List(co..
[...]

Issue #175, opened 2013-06-25
Classpath issue with EAR and Websphere


ghost:
_Original author: marcin.j...@gmail.com (January 28, 2013 19:57:05)_

Hi, I'm trying to use flyway in a multi-module project:

my-project
-ejb
-war
-ear

I've added migration code in ejb module in @Startup and @Singleton bean in @PostConstruct method. Ive put migrations to /src/main/resources/db/migrations in ejb module, but they cannot be found. 

I think its classpath issue - ClassPathScanner.getClassLoader returns EAR's classloader but I need ejb's module classloader. 

Is there other solution than modifying code and passing right classLoader to ClassPathScanner ? Like that :

Flyway f = new Flyway();
f.setClassLoader(this.getClassLoader());
f.migrate();

I use Websphere 8.5.0.1 with JEE6. 

Thanks!

_Original issue: http://code.google.com/p/flyway/issues/detail?id=423_

PR #1301, opened 2016-04-28
Fix IBM classloader regression

Sp0i3eR:
Migration during #979 fix introduced regression on #175
Migration to Location from String in ClassPathResolver on cbc9fef introduced regression in websphere hack from #175 in ClassPathScanner. Location there resolves to "classpath:db/migration/flyway.location" which renders resource not visible for IBM classloader. Added getPath() like in normal classloader flow. Tested on WebSphere 8.5.5.9 jdk8.


axelfontaine:
Thanks! Merged.

Issue #1333, opened 2016-06-05
OracleDB failed to migrate while schema name contains character "-"


eugenpp:
###### What version of Flyway are you using?

4.0.1
###### What database are you using (type & version)?

Oracle 11.2
###### What operating system are you using?

Windows, Linux
###### What did you do?
- run command line "flyway migrate" or "flyway repair" or whatever
- run migration within java application
###### What did you expect to see?

Migrated database.
###### What did you see instead?

Database unchanged.

```
Database: jdbc:oracle:thin:@**** (Oracle 11.2)
....
ERROR: Error restoring current schema to its original setting
```

with -X switch enabled....

```
C:\tmp\flyway-4.0.1>flyway -X repair
Flyway 4.0.1 by Boxfuse

DEBUG: Java 1.6.0_45 (Sun Microsystems Inc.)
DEBUG: Windows 8 6.2 amd64

...
DEBUG: flyway.jarDirs -> C:\tmp\flyway-4.0.1\jars
DEBUG: flyway.password -> *********
DEBUG: flyway.url -> jdbc:oracle:*****************
DEBUG: flyway.user -> MY-SCHEMA
DEBUG: flyway.locations -> classpath:db.migration
...
Database: jdbc:oracle:thin:********************* (Oracle 11.2)
DEBUG: DDL Transactions Supported: false
DEBUG: Schema: "MY-SCHEMA"
DEBUG: Spring Jdbc available: false
DEBUG: Scanning for SQL callbacks ...
DEBUG: Scanning for classpath resources at 'classpath:db/migration' (Prefix: '', Suffix: '.sql')
DEBUG: Determining location urls for classpath:db/migration using ClassLoader sun.misc.Launcher$AppClassLoader@40affc70 ...
...
ERROR: Unexpected error
org.flywaydb.core.api.FlywayException: Error restoring current schema to its original setting
        at org.flywaydb.core.internal.dbsupport.DbSupport.restoreCurrentSchema(DbSupport.java:132)
        at org.flywaydb.core.internal.command.DbRepair.repair(DbRepair.java:144)
        at org.flywaydb.core.Flyway$6.execute(Flyway.java:1100)
        at org.flywaydb.core.Flyway$6.execute(Flyway.java:1098)
        at org.flywaydb.core.Flyway.execute(Flyway.java:1365)
        at org.flywaydb.core.Flyway.repair(Flyway.java:1098)
        at org.flywaydb.commandline.Main.executeOperation(Main.java:145)
[...]

PR #1334, opened 2016-06-05
#1333 OracleDB failed to migrate while schema name contains character "-"

eugenpp:
… "-"


axelfontaine:
Thanks for submitting this! I didn't merge it as we already have a dedicated quote method for that.

PR #1338, opened 2016-06-10
Reorder log detection sequence.

ptahchiev:
If one is using flyway from a maven plugin there is no way to control the log sequence. Simply because flyway checks very early if slf4j is present and maven provides SL4J's `SimpleLogger` out of the box. However to change the `SimpleLogger` log level's one must go to ` ${MAVEN_HOME}/conf/logging/simplelogger.properties` and we can't tell every user to change his maven settings. 

This pull-request gives priority to detecting if commons-logging is available. This way the user can route commons-logging to whatever their logging implementation of choice is (log4j, log4j2, logback, etc.) and be able to control the log level.


axelfontaine:
As commons-logging is effectively legacy, this PR would now again favor it over SLF4J for all users. This is I am afraid not an option.

Additionally to cleanly tie into Maven's logging one MUST use slf4j. (Which is not to say I don't feel your pain. See https://issues.apache.org/jira/browse/MNG-5916)

ptahchiev:
Would it possible to add some more than? Perhaps Log4j2 before the SLF4j?

PR #1339, opened 2016-06-15
SQLite blob literal parsing support

aaltergot:
`SQLiteSqlStatementBuilder` gets stuck in `insideQuoteStringLiteral` state when encounters [sqlite blob literals](https://www.sqlite.org/lang_expr.html) like `x'00'` or `X'00'`.


axelfontaine:
Thanks! Merged.

PR #1340, opened 2016-06-15
SQLite sqlite_sequence no drop on clean

aaltergot:
`Flyway#clean()` fails with
> java.sql.SQLException: [SQLITE_ERROR] SQL error or missing database (table sqlite_sequence may not be dropped)

if schema contains `sqlite_sequence` table.
> [The sqlite_sequence table is created automatically whenever any ordinary table with an AUTOINCREMENT integer primary key is created. Once created, the sqlite_sequence table exists in the sqlite_master table forever; it cannot be dropped.](https://www.sqlite.org/fileformat2.html#seqtab)


aaltergot:
Disregard this one, I've got better idea.

PR #1341, opened 2016-06-16
SQLite system table no drop

aaltergot:
`Flyway#clean()` fails with
> java.sql.SQLException: [SQLITE_ERROR] SQL error or missing database (table sqlite_sequence may not be dropped)

if schema contains `sqlite_sequence` table.
> [The sqlite_sequence table is created automatically whenever any ordinary table with an AUTOINCREMENT integer primary key is created. Once created, the sqlite_sequence table exists in the sqlite_master table forever; it cannot be dropped.](https://www.sqlite.org/fileformat2.html#seqtab)


axelfontaine:
Another great PR! Many thanks! Merged.

axelfontaine:
@aaltergot Could you please get in touch with me via email? (axel at boxfuse.com) Thanks!

Issue #1180, opened 2016-01-24
sqlite: flyway migrate fails, reports database is locked


masc3d:
I'm trying to setup flyway with sqlite, org.xerial:sqlite-jdbc and gradle.
While `baseline` works fine, `migrate` always reports `database is locked`

```
Unable to restore autocommit to original value for connection
java.sql.SQLException: database is locked
    at org.sqlite.core.DB.throwex(DB.java:859)
    at org.sqlite.core.DB.exec(DB.java:142)
    at org.sqlite.jdbc3.JDBC3Connection.setAutoCommit(JDBC3Connection.java:155)
    at org.flywaydb.core.internal.util.jdbc.TransactionTemplate.execute(TransactionTemplate.java:96)
        ...
```

axelfontaine:
How did you set up your datasource? Please include a small test case to reproduce.

masc3d:
sure. happens with the most basic setup. https://github.com/masc3d/flyway-issue-1180

masc3d:
issue persists with flyway/gradle-plugin 4.0. do you require more feedback?

masc3d:
adding a task using flyway directly works pretty much flawlessly.

```
/** masc20160330. workaround for flyway bug https://github.com/flyway/flyway/issues/1180 */
task flywayMigrateWorkaround << {
    Flyway flyway = new Flyway();
    flyway.setDataSource("jdbc:sqlite:${db_file}", "", "");
    flyway.migrate();
}
tasks.flywayMigrateWorkaround.group = "flyway"
```

btw flywayClean fails as well. looks like flyway-gradle-plugin is pretty much incompatible with sqlite.

```
  Unable to drop "main"."sqlite_sequence"
  [SQLITE_ERROR] SQL error or missing database (table sqlite_sequence may not be dropped)
```

axelfontaine:
Both the original and the sqlite_sequence issue have now been fixed. See my comment on #1499

Thanks for investigating!

PR #1342, opened 2016-06-16
Support for external data sources in single connection mode

aaltergot:
Flyway fails to migrate SQLite on provided application level DataSource instance.
> Caused by: java.sql.SQLException: [SQLITE_BUSY]  The database file is locked (database is locked)

As far as it is not possible to define database type through `dataSource` , `setSingleConnectionDataSource(DataSource)` was added to facade to workaround the bug.
Known related issue: #1180


axelfontaine:
Added another configuration setting is not the way to go here. This should work transparently with the regular setting instead.

aaltergot:
@axelfontaine you probably right. It's just came to me that I can feed Flyway with already wrapped datasource (solve the issue on client side). Thank you.

PR #1345, opened 2016-06-20
Make baseline() idempotent even after applying migrations

ferstl:
Baselining a second time after applying migrations threw an Exception even though the baseline version and description is the same. This PR allows executing the baseline command again after applying migrations **as long as the baseline version and description do not change**. In such a case the baseline command won't do anything.
This feature is especially useful in automated test environments where baselining or migrations might have taken place or not.

Before changing the `DbBasline` class, I wrote some unit tests to make sure that I don't accidentally break something. After changing the class I added some more tests for this new feature.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/command/DbBaseline.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/command/DbBaseline.java
@@ -112,9 +112,6 @@ public class DbBaseline {
             new TransactionTemplate(connection).execute(new TransactionCallback<Void>() {
                 public Void doInTransaction() {
                     dbSupport.changeCurrentSchemaTo(schema);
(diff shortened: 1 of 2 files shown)


axelfontaine:
Thanks for the pull request. I personally don't see the exact need for this. However I am all for making things idempotent and the change is rather small, so approved.

ferstl:
@axelfontaine For us it will be useful in test environments where we use Flyway to test db migrations on schemas with (almost) productive data. The test does basically set a baseline and applies new scripts. If the baseline was already set (because the test has already been executed once on that schema), the build crashed. With this fix we can remove the hacks we had to make to work around this issue.

axelfontaine:
@ferstl Can you get in touch via email with me? (axel at boxfuse.com) Thanks!

PR #1356, opened 2016-07-06
Removing redundant validation statement

izzabudaka:


axelfontaine:
Thanks, but no thanks. This breaks the singular form.

Cheers
Axel

PR #1375, opened 2016-07-27
Support dependency injection when resolving JdbcMigrations

stacycurl:
By extracting the instantiation of JdbcMigrations into a protected 'instantiate' method
it's possible for users of flyway to integrate with a dependency injection
framework by overriding 'instantiate' for example

```
Flyway flyway = new Flyway();
flyway.setSkipDefaultResolvers(true);
flyway.setResolvers(new JdbcMigrationResolver(...) {
    @Override
    protected JdbcMigration instantiate(String className, ClassLoader classLoader) throws Exception {
        return ...;
    }
});
```

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/resolver/jdbc/JdbcMigrationResolver.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/resolver/jdbc/JdbcMigrationResolver.java
@@ -82,7 +82,7 @@ public class JdbcMigrationResolver implements MigrationResolver {
         try {
             Class<?>[] classes = scanner.scanForClasses(location, JdbcMigration.class);
             for (Class<?> clazz : classes) {
-                JdbcMigration jdbcMigration = ClassUtils.instantiate(clazz.getName(), scanner.getClassLoader());
+                JdbcMigration jdbcMigration = instantiate(clazz.getName(), scanner.getClassLoader());
                 ConfigurationInjectionUtils.injectFlywayConfiguration(jdbcMigration, configuration);
 
                 ResolvedMigrationImpl migrationInfo = extractMigrationInfo(jdbcMigration);
@@ -97,6 +97,19 @@ public class JdbcMigrationResolver implements MigrationResolver {
 
         Collections.sort(migrations, new ResolvedMigrationComparator());
         return migrations;
+    }
+
+    /**
+     * Creates a new instance of this class.
+     *
+     * @param className   The fully qualified name of the jdbc migration class to instantiate.
+     * @param classLoader The ClassLoader to use.
+     * @return The new instance.
+     * @throws Exception Thrown when the instantiation failed.
+     */
+    @SuppressWarnings({"unchecked"})
+    protected JdbcMigration instantiate(String className, ClassLoader classLoader) throws Exception {
+        return ClassUtils.instantiate(className, classLoader);
     }
 
     /**


stacycurl:
Sorry I only read your contribution guidelines after submitting the PR, too eager to code I guess !
wrt testing: I could add to the JdbcMigrationResolverSmallTest if that would improve this PR, and I don't mind adding a new feature request issue too.

ben-manes:
Since this only adds to the default resolvers, will this work? I'd like to inject into a `JdbcMigration` (at the standard location) using Guice. The Spring solutions use classpath scanning to eagerly inject into statics, both of which are frowned upon in Guice.

But DI does seem a little inappropriate for this stage of the lifecycle, so its really about getting a few helpers to make the migration code a little friendlier. In my case to use jOOQ DDL statements and to use existing sql format settings.

stacycurl:
You can already suppress the default resolvers, I've gotten this to work by copy-and-pasting the entirety of JdbcMigrationResolver and using something other than ClassUtil to instantiate.

What do you mean 'at the standard location', is there some other way to use DI with JdbcMigrations, I don't want to use Spring, or statics, just simple CDI.

I don't know what you mean by inappropriate, do you mean inappropriate to use DI for JdbcMigrations ?

ben-manes:
Oh I wasn't complaining. I was trying to debug how to use Guice and `CompositeMigrationResolver` will always a raw `JdbcMigrationResolver` into the list. This would mean, if I'm not mistaken, the class would be resolved by both it and any custom resolver. So the workaround would be to query a distinct location so that `JdbcMigrationResolver` doesn't pull up those classes by scanning.

In Guice, a common approach is to use the dependency graph to handle some of the startup life-cycle phases. So a `Provider` would be called to obtain a `DataSource`, which would run the migrations before handing it out. This way any consumers are ensured to have a database in the expected state. One might use an explicit lifecycle (e.g. Guava's `ServiceManager`), but have to be careful that 3rd party modules are not eagerly using the connection.

There is also the concern that DI would introduce a lot of complex migration code calling into libraries. Even simple cases like reusing a repository. But then a minor refactoring would cause a checksum change. So keeping the Java migration minimal of dependencies is then a good idea.

In my case I realized the DI need was unnecessary and not beneficial. But I agree with your PR that the class instantiation should be DI friendly as the flexibility could be warranted, even if abused.

ben-manes:
Oh I wasn't complaining. I was trying to debug how to use Guice and `CompositeMigrationResolver` will always add a raw `JdbcMigrationResolver` into the list. This would mean, if I'm not mistaken, the class would be resolved by both it and any custom resolver. So the workaround would be to query a distinct location so that `JdbcMigrationResolver` doesn't pull up those classes by its scanning.

In Guice, a common approach is to use the dependency graph to handle some of the startup life-cycle phases. So a `Provider` would be called to obtain a `DataSource`, which would run the migrations before handing it out. This way any consumers are ensured to have a database in the expected state. One might use an explicit lifecycle (e.g. Guava's `ServiceManager`), but have to be careful that 3rd party modules are not eagerly using the connection.

There is also the concern that DI would introduce a lot of complex migration code calling into libraries. Even simple cases like reusing a repository. But then a minor refactoring would cause a checksum change. So keeping the Java migration minimal of dependencies is then a good idea. That fits into Flyway's mental model of being SQL first.

In my case I realized the DI need was unnecessary and not beneficial, as I could still wire jOOQ without much difficulty. But I agree with your PR that the class instantiation should be DI friendly as the flexibility could be warranted, even if abused.

stacycurl:
I'm sorry if I came across as complaining, that's not my intent, I think flyway is a great library.

Let me provide a little background. My team is using flyway to migrate json data stored in oracle (can't budge company to use a json database) so we extend JdbcMigration to a specialised DealJsonMigration:

trait DealJsonMigration extends JdbcMigration {
  def migrate(connection: Connection): Unit = // query deals table, parse as Json, send each row to migrate(Json)

  def migrate(dealJson: Json): Unit
}

However sometimes we need access to other json data stored in other tables and want something like this:

class V101__TrickyDealMigration(otherJsonStore: OtherJsonStore) extends DealJsonMigation {
  def migrate(dealJson: Json): Unit = ???
}

Without DI we'd have to have a mutable 'otherJsonStore' & any other dependency in DealJsonMigration, it's also trickier to test.



I don't think flyway should have DI itself, that would bias it to a particular flavour of DI and make the code more complex, it's just a pity that in it's current form it's not open to that kind of extension, or is it ? I'm not hugely familiar with it's entire structure & how configurable it is. 

Even with my proposed changes to JdbcMigrationResolver I'd still need global mutable state if I wanted to share injected instances across JdbcMigrations, but it seemed the simplest route to making some form of DI possible.

axelfontaine:

Issue #1382, opened 2016-07-31
Inconsistent usage documentation: `conf/flyway.conf` vs `conf/flyway.properties` vs `./flyway.conf`


sw00:
I'm using flyway as a standalone CLI and found the information with regard to the default `configFile` option to be inconsistent and confusing. As I understand, flyway will look for a `conf/flyway.properties` only if invoked as a Maven plugin and not as a CLI.

```
$ flyway -h
Flyway 4.0.3 by Boxfuse

Usage
=====

flyway [options] command

By default, the configuration will be read from conf/flyway.conf.
Options passed from the command-line override the configuration.
```

```
configFile                   : Config file to use (default: conf/flyway.properties)
```

axelfontaine:
Flyway will look for config files in the order specified here: https://flywaydb.org/documentation/commandline/#configFile

Works as designed.

sw00:
Hi, thanks for looking at this.

In your link, I see no mention of a `flyway.properties` file - only `flyway.conf`. 

If you could, please explain where a `*.properties` file is loaded from when using the CLI - I can find no line where this happens in `org.flywaydb.commandline.Main`.

axelfontaine:
Sorry I was a bit too quick here. Please see my comment on #1383

sw00:
Great, thanks for the clarification. See https://github.com/flyway/flyway/pull/1383

PR #1383, opened 2016-07-31
Replace references to deprecated "flyway.properties" file.

sw00:
Resolves https://github.com/flyway/flyway/issues/1382

According to the following [line](https://github.com/sw00/flyway/blob/0ef790a4a71e91297ce6ff02754dfa203be1fdfb/flyway-commandline/src/main/java/org/flywaydb/commandline/Main.java#L364), flyway will by default load `./flyway.conf` and not `conf/flyway.properties` when executed as a CLI.

### DIFF
--- a/flyway-commandline/src/main/java/org/flywaydb/commandline/Main.java
+++ b/flyway-commandline/src/main/java/org/flywaydb/commandline/Main.java
@@ -254,7 +254,7 @@ public class Main {
         LOG.info("baselineVersion              : Version to tag schema with when executing baseline");
         LOG.info("baselineDescription          : Description to tag schema with when executing baseline");
         LOG.info("baselineOnMigrate            : Baseline on migrate against uninitialized non-empty schema");
-        LOG.info("configFile                   : Config file to use (default: conf/flyway.properties)");
+        LOG.info("configFile                   : Config file to use (default: ./flyway.conf)");
         LOG.info("configFileEncoding           : Encoding of the config file (default: UTF-8)");
         LOG.info("jarDirs                      : Dirs for Jdbc drivers & Java migrations (default: jars)");
         LOG.info("");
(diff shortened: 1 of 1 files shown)


axelfontaine:
This is incorrect. See comment on #1382.

axelfontaine:
Sorry, I was a bit too quick on the draw here. The change from properties to conf took place sometime ago and somehow this reference fell through the cracks. The conf folder however is correct (see comment on #1382)

Can you fix this PR so it can be merged?

flyway-commandline/src/main/java/org/flywaydb/commandline/Main.java
```diff
@@ -254,7 +254,7 @@ private static void printUsage() {
         LOG.info("baselineVersion              : Version to tag schema with when executing baseline");
         LOG.info("baselineDescription          : Description to tag schema with when executing baseline");
         LOG.info("baselineOnMigrate            : Baseline on migrate against uninitialized non-empty schema");
-        LOG.info("configFile                   : Config file to use (default: ./flyway.conf)");
+        LOG.info("configFile                   : Config file to use (default: <install-dir>/conf/flyway.conf)");
```
sw00: I'm specifying the the `<install-dir>`, otherwise it looks like it looks for a `conf/flyway.conf` relative to the working directory you execute the `flyway` CLI in.

Technically, it loads `<install-dir>/conf/flyway.conf`, `<user-home>/flyway.conf` and `<current-dir>/flyway.conf` by default, but I believe the important change is removing mentions of `flyway.properties`

axelfontaine: Much better indeed! Thanks!

axelfontaine:
Merged! Thanks! I really appreciate polishing those details. They are tiny but important to avoid confusion. 👍

PR #1393, opened 2016-08-05
Add the notion of optional migration which does not fail the entire migration process on error

reda-alaoui:
Hello,

I need a way to declare some sql migrations as optional to avoid full migration failure on those kind of SQL errors:
- Trying to add an already existing column to a table
- Trying to add an already existing table to a database
- Trying to insert a duplicate row
- Trying to drop non existing column from a table
- etc ...

To give this option, I propose to change the migration file name pattern.
The pattern would change from `V1_2__Hello_World.sql` to:
- `V1_2__optional__Hello_World.sql`
OR
- `V1_2__required__Hello_World.sql`

Current and new patterns would be accepted to begin.
Current one would be removed in version 6 for example.

The patch I propose is big. I am willing to modify any aspect.

Kind regards


axelfontaine:
Thank you for your contribution. Sorry to disappoint you, but this is not something I plan to add to the core distribution. One of the main guarantees that Flyway offers you, is that if you don't modify your schema behind its back, all migrations _will_ succeed.

reda-alaoui:
At our company, we have thousands of databases using the same schema.
Sometimes, data or structures can be altered by direct terminal manipulation or software bug over time.
When we try to clean those databases, we know that our cleaning scripts can fail.

If the user is aware that optional means no guarantee of success, I don't see why it should be an issue. Aren't flyway end users assumed to be professionals?

Without this patch, the only way to pass those kind of sql scripts is to wrap them into pl/sql functions which catch every exception. This is not satisfying for us (and IMHO for many people).

Perhaps we could add a status between success and failure in the meta table?

PR #1394, opened 2016-08-06
Add an option to append description int hash to the version

reda-alaoui:
In our team of developers, working on multiple feature branches, we frequently end up with conflicts when two developers create migration files with exact same version.

We always resolve those conflicts by incrementing the version of the last commited file. That means, the order between those conflicted updates does not matter because they are almost always unlinked.

I think the easiest solution is to automatically generate the minor part of the version number by computing a hash on the description. This hash would be then appended to the base version.

Example:
- V2_1__john.sql => 2.1.1346565
- V2_1__jane.sql => 2.1.1347966

Regards

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
@@ -296,6 +296,11 @@ public class Flyway implements FlywayConfiguration {
     private boolean dbConnectionInfoPrinted;
 
     /**
+     * Whether description hash should be appended to the version
+     */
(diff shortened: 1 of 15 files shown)


axelfontaine:
If you are not dependent on the order, then the outOfOrder flag combined with reverse timestamp version (20160808080356) already offer a great solution today. The proposed implementation is not something I see as belonging in the main distribution. You can however easily use this as an extension by wrapping it in your own MigrationResolver.

reda-alaoui:
Ok we will use our issue tracker generated numbers to make the distinction.

Issue #897, opened 2014-12-02
Allow custom migration resolver to access SQL migration executor


ataastar:
Hi!

I apologize if it's a solved a problem
we have fixes script structure. Example:
20141201/1_update.sql
20141201/2_update.sql
20141202/1_update.sql
..
1. So we need the possibility to resolve migration info other way instead of MigrationInfoHelper
Or
2. override SqlMigrationResolver or implement MigrationResolver
I look after to implement custom migration resolver, but it can't access the parameterized objects (dbSupport, locations, ...)
Is there a way to override SqlMigrationResolver or implement MigrationResolver and access the configured object? 
I'm sorry but I could not find any examples

Thanks in advance,
Attila

DoodleBobBuffPants:
Solved as we allow providing your own [MigrationResolver](https://flywaydb.org/documentation/configuration/parameters/resolvers)

PR #1395, opened 2016-08-06
Add the ability to set a custom ResolvedMigration comparator

reda-alaoui:


axelfontaine:
This are low-level details that should be handled using a custom MigrationResolver and a custom ResolvedMigration class. I do not want to expose this kind of property at this point.

reda-alaoui:
I proposed this PR because it is impossible to just extend an already existing MigrationResolver such as the SqlMigrationResolver.

One has to copy resolver and executor to start customization.
The issue has been raised in #897 in Dec 2014, nothing since.

But I assume there is no hope for this PR.

PR #1397, opened 2016-08-07
Support multiple migrations per transaction

### LINKED ISSUE #181: Support Multiple Migrations within a single Transaction
_Original author: ke...@fidelma.com (April 24, 2013 05:01:42)_

When upgrading a database I would expect all un-applied migrations to be applied within the same transaction. Currently each migration runs within its own transaction.

My use-case looks something like this:
1. Version 1.0 is deployed into Production
2. Developers work on features for Version 2.0, each creating a migration script matching their new feature.
3. As development progresses, V2.x migrations, and the associated application, are put into QA.
4. Eventually we decide we've done enough, and want to release v2, and it's myriad migration scripts.
5. When we run the deployment against v1 in production we expect all the v2.x migrations to run, and if there is a problem with any of them, then the database is rolled back, and we revert the application to v1 while we figure out what went wrong. 

We really don't want _some_ of the v2 scripts to be committed to the production database.

We're using Postgres, so DDL rolls back nicely.

I'm happy to do the change, if nobody else is interested, but it seems like an &quot;obvious&quot; feature to me. None of your competition seem to offer it either. Am I missing something?

_Original issue: http://code.google.com/p/flyway/issues/detail?id=484_

reda-alaoui:
Fixes #181 .

Added a notion of batch of migrations.
A batch of migrations executes into a single transaction.
By default a batch of migrations consists of only one migration.

To modify batch creation, one have to implement the MigrationBatchService and pass it to Flyway object:

```java
public interface MigrationBatchService {
    /**
     * @param dbSupport
     * @param migrationInfo
     * @return True if the migration is to be considered as the last of its batch.<br>
     *      A migration marked as last of batch will be followed by a commit on success.
     */
    boolean isLastOfBatch(DbSupport dbSupport, MigrationInfo migrationInfo);
}
```

Therefore we can have the following behaviours:
- One migration per transaction (default)
- All migrations into one transaction
- Migrations grouped by major versions

Regards

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
@@ -22,6 +22,8 @@ import org.flywaydb.core.api.MigrationVersion;
 import org.flywaydb.core.api.callback.FlywayCallback;
 import org.flywaydb.core.api.configuration.FlywayConfiguration;
 import org.flywaydb.core.api.resolver.MigrationResolver;
+import org.flywaydb.core.internal.batch.DefaultMigrationBatchService;
+import org.flywaydb.core.internal.batch.MigrationBatchService;
 import org.flywaydb.core.internal.callback.SqlScriptFlywayCallback;
 import org.flywaydb.core.internal.command.DbBaseline;
 import org.flywaydb.core.internal.command.DbClean;
@@ -296,6 +298,11 @@ public class Flyway implements FlywayConfiguration {
     private boolean dbConnectionInfoPrinted;
 
     /**
+     * The migration batch service to use
+     */
+    private MigrationBatchService migrationBatchService = new DefaultMigrationBatchService();
+
+    /**
      * Creates a new instance of Flyway. This is your starting point.
      */
     public Flyway() {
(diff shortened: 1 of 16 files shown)


axelfontaine:
Thank you for your contribution. If we were to introduce this feature, we wouldn't do it with a new concept like this. Instead we'd try to make it as transparent as possible to the end user, make sure it works on all types of clients (not just the Java API), be well documented, and not expose any private API in the public interface.

I realize the rejections of your contributions may seem like a disappointment. As we have stated in the [contribution guide](https://flywaydb.org/documentation/contribute/code/submit):

> The first step is to raise an issue in the Issue Tracker. This issue will be the place where you will be able to communicate with Flyway's committers and other users. You can use it to present your ideas and gather feedback.

We really mean this, as we don't mind you to feel like you've invested a lot of time and effort in something that has zero chance of ever being merged.

reda-alaoui:
Hello Axel,

The issue #181 was created in 2013.

Since then, there have been a lot of discussion about it, and the only conclusion I understood was that you were waiting for sponsorship or someone's else PR.

Nevertheless, I assumed we could talk of the patch code in the PR, not in the issue.

I don't totally understand what you mean by `Not just the Java API`. Today, it seems that `flyway-core` is only focused on the JDBC api?

Issue #1396, opened 2016-08-07
Log the inner exception thrown from JdbcMigrationResolver.resolveMigrations()


logicbomb:
###### What version of Flyway are you using?

4.03
###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin, SBT plugin, ANT tasks)

Command-line
###### What database are you using (type & version)?

Postgres 9.4
###### What operating system are you using?

Ubuntu 14.04
###### What did you do?

(Please include the content causing the issue, any relevant configuration settings, and the command you ran)
Attempted to execute a java migration
###### What did you expect to see?

Helpful error message
###### What did you see instead?

> ERROR:  Unable to resolve Jdbc Java migrations in location: classpath:db/migration

It would be nice if flyway logged the inner exception thrown from `JdbcMigrationResolver.resolveMigrations()`  I'm sure a lot of Java migration errors are a result of the java implementation and nothing to do with Flyway, but as it stands now it's tough to have any idea what's going on.  Even changing the error message to include (something like) "try re-running the migration with the -X switch to see detailed logs" would be a big help.

PR #1400, opened 2016-08-09
ISSUE 1396: Invite users to use -X switch

logicbomb:
A very simple PR to address issue #1396

### DIFF
--- a/flyway-commandline/src/main/java/org/flywaydb/commandline/Main.java
+++ b/flyway-commandline/src/main/java/org/flywaydb/commandline/Main.java
@@ -106,7 +106,9 @@ public class Main {
                 LOG.error("Unexpected error", e);
             } else {
                 if (e instanceof FlywayException) {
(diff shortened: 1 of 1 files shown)


axelfontaine:
This is not the correct place to fix this as -X means nothing to API users for example. This must be addressed generically in the command-line module.

logicbomb:
Oh yes, I see now.  I think this is applicable to all exceptions that are caught by the command line, so I think it makes sense to simply modify `/Users/jason/src/flyway/flyway-commandline/src/main/java/org/flywaydb/commandline/Main.java` Line 108, does that work for you?

axelfontaine:
Yes

logicbomb:
updated

axelfontaine:
Upon second though, I am not entirely happy both with the wording of the message as well as the need for it. For #1396 fixing the original message is probably the better approach.

Issue #1403, opened 2016-08-11
Flyway migration ordering


tfsoares:
##### Environment
I'm using Ubuntu 16.04, PostgreSQL 9.4 and the Java client.

##### Problem
Run repeatable migrations that depend on another repeatable migration. 

##### Explanation
This brings a problem because the implicit sorting is alphabetical and I needed to order them differently.

Imagine I have this 2 migrations:
- `db/migrations/R__vw_a.sql`
- `db/migrations/R__vw_b.sql`

The current sorting would lead to `vw_a` to be ran before `vw_b`. The problem in this is that `vw_a` might perform a query on `vw_b`. So migrating `vw_a` before `vw_b` will generate an error and the migration will never succeed.

So a Custom Migration Resolver might be used to retrieve a list of Resolved Migrations with the correct sorting.
But at Flyway current state, after using the Custom Resolver, it will reorder migrations again.

##### Possible Solution
One solution might be add a property to set wether to use or to skip default sorting. This property would be used by an if condition on Composite Resolver.
Then, on MigrationInfoServiceImpl, change some collections from Unordered to Ordered types and tweak MigrationInfoImpl's compareTo method to return 0 when comparing Repeatable Migrations.

This way it would be possible to keep the ordering retrieved from Resolvers.

axelfontaine:
In your case the simplest this is probably to combine both sql files into 1 or look at versioned migrations instead for those objects.

I currently don't feel that adding an additional flag here would carry its additional conceptual weight.

albertocsm:
I don't agree with you @axelfontaine.
Combining those scripts in just one file doesn't make sense to me. Separate objects should have their own life cycle and evolve separately.
Besides,  the object scripted in 'a.sql' can be used in multiple other views ('b.sql', 'c.sql', ect) and scripting those dependencies all over the place is tremendously difficult to maintain. 

On the other hand, versioning stateless objects with 'versioned migrations', IMHO, its not a good way to keep track of the changes. We followed that road in @Bikeemotion and you can get lost pretty quickly after a few iterations.  
SVC can handle stateless object's repeatable scripts a lot better and help me maintain whats left of my sanity!

alkamo:
The proposed solution only works if the two views were created as part of separate releases. If I were trying to solve this problem, I think I'd resort to adding a sequence number as a prefix in the file name, in order to ensure the correct ordering if two related views are added as part of the same release
```
db/migrations/R__002_vw_a.sql
[...]

PR #1405, opened 2016-08-11
Flyway Repeatable Migration ordering

tfsoares:
Refers to #1403


maxsoelalberton:
@tfsoares, I have a problem with a repeatable migration, when I revert a change from a script.
Example:
I had a script with the checksum 353842862.
I made a change in this script, and the checksum was changed to 1106852894.
After that, the change was reverted, and the checksum was back to 353842862.
When I runned Flyway again, this script wasn't considered in migration, because that checksum already existed in database, even not being the last migration.
Your change resolve this situation?

axelfontaine:
Closing. See my comments on #1403

PR #1406, opened 2016-08-12
Add the ability to dynamically ignore some sql errors on sql migration's scripts

### LINKED ISSUE #1175: Need a way to ignore ORA errors and continue executing other stmts in a file
Very often developers , would fix simple uniq constraint violation kind of errors , offline / after deployment. Having 1000s of insert/update stmts in set of files, if flyway stops on encountering error and refuse to execute remaining stmts in file, we feel it creates delay in prod deployment ( with limited window time). Need a flag ( like -X for extended debug ) , when we set, flyway could continue executing other stmts in a file.  it will be better , if we have a way to specify what kind of error codes can be ignored. like a separate file ( like callback) for ignorable error codes.

As of now, we have work around to completely skip one file ( by changing its prefix/suffix). THere is no way to skip erring stmts in a file.

This will help dev/ops teams to complete prod deployment with in scheduled time, and CI/CD will be easy.

reda-alaoui:
Fixes #1175

I don't think it is a niche requirement.
In many DB provider such as PostgreSQL, there is no way to check for some object existence before creating them.
For example you cannot know if a column already exists before creating in one simple SQL line.

Moreover, SqlMigrationResolver being still impossible to extend, I think this feature is required.

Regards

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
@@ -21,6 +21,7 @@ import org.flywaydb.core.api.MigrationInfoService;
 import org.flywaydb.core.api.MigrationVersion;
 import org.flywaydb.core.api.callback.FlywayCallback;
 import org.flywaydb.core.api.configuration.FlywayConfiguration;
+import org.flywaydb.core.api.migration.sql.SqlMigrationScriptExecutionInterceptor;
 import org.flywaydb.core.api.resolver.MigrationResolver;
 import org.flywaydb.core.internal.callback.SqlScriptFlywayCallback;
 import org.flywaydb.core.internal.command.DbBaseline;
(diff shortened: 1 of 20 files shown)


bladeoflight16:
> In many DB provider such as PostgreSQL, there is no way to check for some object existence before creating them.

This assertion is *dead* wrong. I can think of three ways of handling this off hand:

  * 9.6 adds the `IF NOT EXISTS` clause to [`ADD COLUMN`](https://www.postgresql.org/docs/9.6/static/sql-altertable.html).
  * Prior to 9.6, you can *easily* consult the system views:
    ```
    DO $$
    BEGIN
        IF EXISTS(SELECT * FROM information_schema.columns WHERE table_schema = 'public' AND table_name = 'my_table' AND column_name = 'blah') THEN
            RAISE NOTICE 'Yes';
        ELSE
            RAISE NOTICE 'No';
        END IF;
    END
    $$
    ```

  * You could also just catch the exception:

    ```
    DO $$
    BEGIN
        ALTER TABLE crop_type ADD COLUMN crop_type_id INTEGER;
    EXCEPTION
        WHEN duplicate_column THEN RAISE NOTICE 'Column already exists';
    END
    $$
    ```

I really don't like the idea of this feature. Most DBs (all supported DBs?) make it possible in some way or another to catch specific errors in the procedural language (PL/pgSQL, PL/SQL, T-SQL), and those mechanisms, examining catalog tables to make checks, or things like `IF NOT EXISTS` are better suited for handling these errors than Flyway.

axelfontaine:
I fully agree with @bladeoflight16 .

Won't fix.

PR #1407, opened 2016-08-19
FIX: $user now replaced by username rather than empty string (postgres).

capocannoniere:
By default, postgres configures such a search path for database roles:
"$user",public
To determine database schema, Flyway removes "$user" from a search path and takes the first schema from the remaining list.
This pull request replaces "$user" with a real user name (obtained from "SELECT current_user" query). So, by default, if no schema specified for a Flyway instance, $user schema will be used rather than public.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/dbsupport/postgresql/PostgreSQLDbSupport.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/dbsupport/postgresql/PostgreSQLDbSupport.java
@@ -61,7 +61,15 @@ public class PostgreSQLDbSupport extends DbSupport {
     }
 
     /* private -> testing */ String getFirstSchemaFromSearchPath(String searchPath) {
-        String result = searchPath.replace(doQuote("$user"), "").trim();
+		String currentUser;
+		try {
+			currentUser = jdbcTemplate.queryForString("SELECT " + getCurrentUserFunction());
+		}
+		catch (SQLException ex) {
+			currentUser = "";
+		}
+		
+        String result = searchPath.replace(doQuote("$user"), currentUser).trim();
         if (result.startsWith(",")) {
             result = result.substring(1);
         }


axelfontaine:
What's the motivation for this? Very often a schema with this name doesn't exist and this then leads to errors.

capocannoniere:
Let the database name is "test", and database user name is "john". By default, when creating a new database on a PostgreSQL, it has only "public" schema, and all users have access to it, which is insecure. So, when someone executes SQL like "CREATE TABLE t1(id int);", the table t1 is created in "public" schema, and this table is available to any PostgreSQL user by default.
If a schema named "john" is created, then the situation is different - if we execute SQL "CREATE TABLE t2(id int);", t2 table is created in "john" schema (assuming that someone didn't changed the default search path). And, despite all tables may have been created in "john" schema, Flyway still creates its "schema_version" tables in "public" schema - is it ok?
I think that's the code that determines which schema to use should get the search path from PostgreSQL (maybe by executing "SHOW search_path"), replace $user with username and for all the schemas in this list check if the schema exists. If the first schema exists, then use it, otherwise check the next schema, and so on. If no schemas found, throw an error. Do you think it's a good idea to implement that in PostgreSQLDbSupport class, or some high-level code should do that?

axelfontaine:
Flyway creates the `schema_version` table in the first schema specified by the `flyway.schemas` property. This is an easy workaround for your situation.

capocannoniere:
Yes, I know about Flyway.setSchemas method, but this requires manually specifying the schema name. Thus, changing database user name in the configuration file also requires changing the schema name, and so is for JDBC connection string, as it can contain schema name. So, specifying something like

```
hibernate.connection.url=jdbc:postgresql://localhost:5432/energy
hibernate.connection.username=energy
hibernate.connection.password=
```

is not enough, and one _must_ add something like

`schema=energy`

And if one specified the schema name in JDBC connection string like that:

```
hibernate.connection.url=jdbc:postgresql://localhost:5432/energy?searchpath=otherschema
hibernate.connection.username=energy
hibernate.connection.password=
```

then

`schema=otherschema`

is required. That means extra care about schema name, which is not good.

axelfontaine:
We ended up removing the search_path parsing logic altogether and replaced it with `select current_schema`.

Issue #970, opened 2015-03-12
Android migration scripts on classpath are invisible


lstrzelecki:
I have android project with two modules:
`app` (apk)
`core` (jar)
Module `apk` has dependency to `core` which has on classpath migration script (`core/src/main/resources/db/migration/V1__Create_person_table.sql`).

On main Activity I call 

``` java
SQLiteDatabase db = openOrCreateDatabase(DB_NAME, 0, null);
ContextHolder.setContext(this);
Flyway flyway = new Flyway();
flyway.setDataSource("jdbc:sqlite:" + db.getPath(), "", "");
flyway.setLocations("classpath:db/migration");
flyway.migrate();
```

But migration script wasn't executed. Of course with assets it works.

Here is project which shows whole configuration: https://github.com/issue-example/flywaydb/tree/issue/android_migration_script_on_classpath_jar

axelfontaine:
Thanks for the report. Would you like to contribute a pull request?

Cheers
Axel

lstrzelecki:
ill try in near future

axelfontaine:
Any news?

SStorm:
+1 very obnoxious misfeature!

MikielAgutu:
Blocked by https://github.com/SQLDroid/SQLDroid/pull/91. SQLDroid's JDBC driver needs to be compatible with Flyway before we look at fixing this.

MikielAgutu:
https://github.com/SQLDroid/SQLDroid/pull/91 has now been merged. When they release we can confirm the fix works pick this up again.

DoodleBobBuffPants:
In Flyway V8 we will be removing support for Android, so closing this issue

More information [here](https://github.com/flyway/flyway/issues/3247)

PR #1416, opened 2016-09-01
Searches all dex files for migrations

TomRK1089:
While most apps will have a single dex file with all classes, some applications will leverage the MultiDex solution provided by the support library (and on later VMs supported natively). We need to scan these secondary dex files for migrations as well, as most authors will not manually ensure that their Java migrations are located in the "primary" dex file.


axelfontaine:
Thanks for the pull request. This seams to be creating a bunch of temp files. Are they automatically being cleaned up as well?

TomRK1089:
@axelfontaine good point. The files are being written into the application's cache directory but we ought to clean them up after the classes themselves are scanned and loaded. I can make some changes.

axelfontaine:
Can you clean this up so it can be merged?

TomRK1089:
@axelfontaine can do.

TomRK1089:
@axelfontaine This may be a world record for stale PRs, but I've rebased and updated these changes to work with Flyway 5.0, as well as added the temp file cleanup previously mentioned. The change had dropped off of our radar since the project that ran into this entered maintenance mode.

MikielAgutu:
Is this still relevant? As I understand Flyway's Android support is in a bad way due to a third party driver being broken (https://github.com/flyway/flyway/issues/970#issuecomment-542695204)

DoodleBobBuffPants:
Closing until further activity

Issue #87, opened 2013-06-25
Output to sql file instead of commiting against the DB


flyway:
_Original author: sylvain....@gmail.com (March 18, 2013 05:59:16)_

HI,

We're just starting to use flyway in our project and so far it looks really good. It's gonna help us a lot automatically upgrade all of our different environments.

One thing that would be nice is way to combine all or specific sql versions into one sql file. We would use this because we can't tell our client's IT department (big client) to use flyway for upgrading the prod database and we obviously don't have access to that database.
Currently the only way we can get the database upgraded is by sending them an SQL file (which they run on their stage before prod).

In theory we could manually combine all the relevant files ourselves but that could be a bit cumbersome and error-prone. Also things like placeholders would need to be replaced manually (search &amp; replace but still) with the proper value, again error-prone.

We use maven so a new goal should do it.

That said thanks for the great work!

_Original issue: http://code.google.com/p/flyway/issues/detail?id=452_

b-long:
+1

vbazagad:
+1

neylorsousa:
+1

alxmrtn:
+1

alexeev:
+1

facuerregue:
+1

vvitayau:
+1

jaytradertee:
+1

advancedgithub:
+1

advancedatolompoiko:
+1

This would be the only missing feature that would allow us to use it in production

johnhindley96:
+1

svoeller99:
+1

mgamer:
+1

tihomirsmudj:
+1

ThiyaguE:
+1, This is the most wanted one for a fortune 30 telecom company, to use flyway in enterprise level production systems !

axelfontaine:
@ThiyaguE Feel free to get in touch by email about [sponsoring this issue](http://flywaydb.org/support/#sponsor) for Flyway 4.1 ...

carllindelof:
+1

lijo-jacob:
Any medium to large business would definitely need this feature. Enterprise DBA teams are not going to let flyway migrate the production databases (at least any time soon). They will have to execute the migrations themselves in production. 

+1

lijo-jacob:
+1

kerryland:
On 21 May 2016 at 08:09, Lijo Jacob <notifications@github.com> wrote:

[...]

PR #1454, opened 2016-10-20
WIP: #87; Adding a 'dryRun' feature to Flyway for supported DB types

trumpetx:
I'm still setting up test environments per the contribution guide to add tests for Postgres and other 'supported' databases, but I wanted to get this Work in Progress up for comment.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
@@ -31,6 +31,7 @@ import org.flywaydb.core.internal.command.DbSchemas;
 import org.flywaydb.core.internal.command.DbValidate;
 import org.flywaydb.core.internal.dbsupport.DbSupport;
 import org.flywaydb.core.internal.dbsupport.DbSupportFactory;
+import org.flywaydb.core.internal.dbsupport.DryRun;
 import org.flywaydb.core.internal.dbsupport.Schema;
 import org.flywaydb.core.internal.info.MigrationInfoServiceImpl;
 import org.flywaydb.core.internal.metadatatable.MetaDataTable;
@@ -200,6 +201,11 @@ public class Flyway implements FlywayConfiguration {
      * Whether to automatically call validate or not when running migrate. (default: {@code true})
      */
     private boolean validateOnMigrate = true;
+
+    /**
+     * Whether to perform a Dry Run. (default: {@code false})
+     */
+    private boolean dryRun = false;
 
     /**
      * Whether to automatically call clean or not when a validation error occurs. (default: {@code false})
@@ -419,6 +425,15 @@ public class Flyway implements FlywayConfiguration {
     }
 
     /**
+     * Whether to perform a dry run of the migration
+     *
+     * @return {@code true} if the migration should roll back entirely when complete {@code false} if not. (default: {@code false})
+     */
+    public boolean isDryRun() {
+        return dryRun;
+    }
+
+    /**
      * Whether to automatically call clean or not when a validation error occurs.
      * <p> This is exclusively intended as a convenience for development. Even tough we
      * strongly recommend not to change migration scripts once they have been checked into SCM and run, this provides a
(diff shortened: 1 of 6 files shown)


flyway-core/src/main/java/org/flywaydb/core/Flyway.java
```diff
@@ -906,6 +930,23 @@ public void setSkipDefaultResolvers(boolean skipDefaultResolvers) {
...
+     *  flyway.migrate();
+     * </code></pre>
+     *
+     * @return The number of migrations that would have been successfully applied
+     * @throws FlywayException when the migration would have failed or Dry-run is not supported by the database driver
+     */
+    public int dryRun() throws FlywayException {
+        setDryRun(true);
+        return migrate();
+    }
```
trumpetx: Any command is 'dryrun'able; however, one sees 'migrate' as the obvious candidate for what a 'dry run' is.  I suppose there could be a dryrun of a repair, for example which would be accomplished by:

```java
flyway.setDryRun(true);
flyway.repair();
```

aldas:
just looked at the code and saw that you are using rollback. Do you know that for example you cannot rollback DML statements in Oracle database?

aldas:
and same goes for Postgres

[http://stackoverflow.com/questions/13462475/is-it-possible-to-wrap-ddl-changes-in-a-transaction-in-postgresql](url)

> ALTER TYPE ... ADD VALUE (the form that adds a new value to an enum type) cannot be executed inside a transaction block.

etc remarks

trumpetx:
Yes, @aldas, in this case 'dryrun' is only usable with supported databases.  As for postgres, I've added this feature to support our Postgres databases which seems to work just fine.  I need to do additional testing.

flyway-core/src/main/java/org/flywaydb/core/internal/dbsupport/DryRun.java
```diff
@@ -0,0 +1,77 @@
...
+    private static final String SAVEPOINT_MARKER = "dryrun";
+    private final Map<Connection, Savepoint> dryRunSavepoints = new WeakHashMap<Connection, Savepoint>();
+    private final int connectionCount;
+    
+    public DryRun(int transactionIsolationLevel, DbSupport dbSupport, Connection... connections) {
+        if (dbSupport == null) {
+            throw new NullPointerException("A Dry-run requires dbSupport to determine whether connections can support a rollback successfully.");
+        }
+
+        if (!dbSupport.supportsDdlTransactions()) {
```
trumpetx: @aldas - see validation here... I actually had a block of code which would set up a rollback point and then test rollbacking (catching exceptions) which validated driver that way; however, it looks like Flyway folks had already done this testing on a per db type as supportsDdlTransactions is set on the support object.

If might be good to re-add that validation in addition to the dbSupport value though.  In case a db is configured to not support the transactions or something like that.

aldas: my bad. I only read the code diagonally. I have used this kind of "dryrun" methods in database layer integration tests - before every test transaction is opened and rollback after test ends but using this kind of feature for migrations may have pretty noticeable side effects when used with certain database statements.  For example Postgres manual says

> TRUNCATE is not MVCC-safe. After truncation, the table will appear empty to concurrent transactions, if they are using a snapshot taken before the truncation occurred. See Section 13.5 for more details.
[https://www.postgresql.org/docs/current/static/sql-truncate.html](url)

trumpetx: Truncate IS, however, DDL Transaction safe, just not MVCC safe.  Considering we're not executing concurrent transactions for doing a dry-run of a migration, I don't see how that's a problem.  If there was a concurrent transaction (like we're doing a dry-run on an existing database that's connected to a running applicaiton) that would be a side effect which should be noted in any documentation.  But since nothing can really be done about it other than "Don't do that" - I'm not sure I see a big issue with the implementation.

Do you have an idea how to get around such issues?  This feature is basically what's preventing my company (and many others) from using a neat tool like Flyway in production.  Assuming the known caveats are acceptable, I wouldn't want to just not implement a feature because it's not perfect in all situations.

aldas: that is why there are test and prelive environments. going straight from dev to production is cowboy programming. At the moment you are trying to implement feature as a workaround for flaws in your development process (not having proper test and prelive environments to safely test releases). 

but not everything is bad - I would use dry run to test my migrations only in development. It may be good/faster way to flush out syntax errors that I occasionally make.

trumpetx:
See new PR: https://github.com/flyway/flyway/pull/1458

PR #1458, opened 2016-10-29
#87; Adding a 'dryRun' feature to Flyway for supported DB types

trumpetx:
Changes made to remove conflicts, squished commits. See https://github.com/flyway/flyway/pull/1454 for old pull request.

Changes from last pull request: removed "dryrun" as a target on the command line and as a method on Flyway.java.  Since anything can be a dry run (including baseline), it made sense to just use the config property to set a given command as a dry run; also, to alleviate concerns from @aldas that this may not be appropriate for all users.  As such, individuals wanting to use dryRun from the command line will have to set the property (and hopefully read the documentation).

A side note:

PS - I am completely unable to run the full test suite.  I tried.  No really, I tried a hell of a lot.  I was able to alter the pom.xml enough to run my tests, but I couldn't run all of the others... I kept getting AWS errors and google was unhelpful in getting past them all.  Before you say anything, yes, I tried "-P-InstallableDBTest -P-CommercialDBTest" and the many derivations found in google searches.

PPS - I'll happily write documentation for the feature, but I haven't yet learned how to submit that.  I'll go read the submission guidelines for documentation next.

### DIFF
--- a/flyway-commandline/src/main/java/org/flywaydb/commandline/Main.java
+++ b/flyway-commandline/src/main/java/org/flywaydb/commandline/Main.java
@@ -254,6 +254,7 @@ public class Main {
         LOG.info("configFile                   : Config file to use (default: conf/flyway.properties)");
         LOG.info("configFileEncoding           : Encoding of the config file (default: UTF-8)");
         LOG.info("jarDirs                      : Dirs for Jdbc drivers & Java migrations (default: jars)");
+        LOG.info("dryRun                       : Executes the given command in an isolated transaction and rolls back at the end regardless of the outcome (default: false)");
         LOG.info("");
         LOG.info("Add -X to print debug output");
         LOG.info("Add -q to suppress all output, except for errors and warnings");
--- a/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
@@ -52,6 +52,7 @@ import org.flywaydb.core.internal.util.scanner.Scanner;
 import javax.sql.DataSource;
 import java.sql.Connection;
 import java.sql.SQLException;
+import java.sql.Savepoint;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -76,7 +77,7 @@ public class Flyway implements FlywayConfiguration {
      * Property name prefix for placeholders that are configured through properties.
      */
     private static final String PLACEHOLDERS_PROPERTY_PREFIX = "flyway.placeholders.";
-
+    
     /**
      * The locations to scan recursively for migrations.
      * <p/>
@@ -200,6 +201,16 @@ public class Flyway implements FlywayConfiguration {
      * Whether to automatically call validate or not when running migrate. (default: {@code true})
      */
     private boolean validateOnMigrate = true;
+
+    /**
+     * Whether to perform a dry-run. (default: {@code false})
+     */
+    private boolean dryRun = false;
+    
+    /**
+     * The isolation level of the dry-run.  (default: TRANSACTION_READ_UNCOMMITTED (1) )
+     */
+    private int dryRunTransactionIsolationLevel = Connection.TRANSACTION_READ_UNCOMMITTED;
 
     /**
      * Whether to automatically call clean or not when a validation error occurs. (default: {@code false})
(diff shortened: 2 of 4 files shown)


axelfontaine:
Given how few databases would actually work with this, I am extremely wary to integrate this.

trumpetx:
Is there a way this could be done a different way that I'm not familiar with?  The business need for this is real, and I have no real desire to maintain a fork of Flyway forever.  I'd also rather not use a different DB migration tool.

axelfontaine:
Could you describe your exact business need in a little more detail?

trumpetx:
Well, for starters we use Postgresql -- I couldn't figure out how to get your test harness up in a timely manner hence the limited test cases submitted.  I didn't really check the compatibility with other vendors -- again, test harness issue.  I'd be willing to got the extra mile if that makes a difference...

We deliver an enterprise solution to a few hundred companies so it's not a simple DEV->TST->STG->PRD deployment model for traditional SASS applications.  Because this is an enterprise solution and customers are often given features in a pay-to-play system, we end up with database issues with feature X + Y + Z conflicting with feature B, but not feature A unless feature M is also enabled.  It's sort of silly, but it happens (rarely).  When this happened under our pre-Flyway homegrown upgrade system, the DB insert/update/whatever would fail and the upgrade would simply continue.  Flyway doesn't do that, it fails the DB migration right there.  There's a flag to continue, but the upgrade file itself won't try the additional statements.  

Old way, we'd have a piece of data to fix, but the system itself would 99/100 be up and running within the upgrade window and the single SQL statement could be addressed by a developer in the morning.  Flyway failures make the system completely unusable because the whole migration fails.  This forces a full rollback of the upgrade.  From a business perspective, this is unacceptable.  We work hard to schedule as little downtime for our customers.  We have some customers who have been up and running for multiple years without downtime often with us begging to bring them down for an upgrade.  They're never happy about it and when the downtime is limited to 15 minutes, they usually will allow it.  Forcing a full rollback and scheduling new downtime is not cool at that point.

I totally see why this happens with Flyway vs. a .sql file piped to psql homegrown system.  I'm not fighting the "flyway way", but the problem is solved by a "dry-run" where we can execute an upgrade during our window and first perform a "dry-run" of the upgrade.  If the db upgrade fails, we spit out a message and the support technician doesn't need to now roll back a database after causing a 15 minute period of downtime (which is actually more disruptive than it sounds since it has to do with network connectivity).  So, we're left with the situation of rescheduling, but with no actual negative effects to the customer.

That's why 'dry-run'.  And yes, I'm aware of about 20 other ways to skin the cat through business process changes.  There's the "upgrade it ahead of time" (DEV-TST-STG-PRD method) which is how I'd do it if it was my business (and we do it this way for some larger customers); however, that's a big bit of overhead for us to impose on our customers.  Smaller groups with more limited budgets don't always go for that sort of process.  You might argue that it's their fault if they don't do it that way, but when it fails and it's our fault - it doesn't matter if it could have been prevented by doing it the "right" way, it's still our fault.  Not cool from a business delivery perspective.

I hope that sheds some light.  And, like I said, if you know of a way to accomplish this in a different way (aka, non-transactional which may be supported by more databases), I'm all ears and eyes.

axelfontaine:
Since you obviously know which features have been bought by which customers (otherwise you wouldn't be able to bill them), couldn't these constellations simply be simulated in-house before delivery as part of the regular automated test suite?

trumpetx:
We actually started doing that as part of our "migrate to Flyway" methodology.  This works for a lot of customers, but there are some that do not allow us to maintain backups/copies in-house.  (Gov't, healthcare, etc)  It's small enough that we'd ...probably... be okay, but it's large enough that we don't want it to be the only defense against that one developer joining two tables expecting them to join on column X when in fact it isn't a 1-1 relationship and he gets > 1 row back on the join when feature M is enabled.  (that's what killed us the last time this happened -- to the point where we needed the actual customer's data to replicate it... existing tests just didn't have that permutation -- and enough data -- to hit that)

trumpetx:
Like I said, there are other ways to fix the business problem at hand; however, they're mostly reliant on doing something else where this happens as part of the natural upgrade process:

do upgrade-with-dryrun;
if(exitCode != 0) exit;
else continue;

No need to maintain that big permutation of features in tests (we try, but it's basically impossible to cover it 100%).  It just gets done as-needed.

reda-alaoui:
If I understand well and since you are using Postgres, you could pass to flyway a datasource providing non commitable jdbc connection. Therefore all your migration scripts would execute inside one unique transaction. 

If one script fail, all modifications are rollbacked.
If no error occurs, you would commit the transaction yourself.

That's what we do.

trumpetx:
Very interesting idea - I hadn't considered that @reda-alaoui.  Thanks for the suggestion.  I was hoping to use Flyway from the command line, but we ended up having to create a database-upgrade.jar which was a wrapper around flyway due to "migrating old databases to flyway" reasons.  Since we already have that, a protected datasource would not be difficult to do.  Thanks!

PR #1465, opened 2016-11-14
Add hashCode & equals for FileSystemResource

evoevodin:
Add  hashCode & equals implementation to the `FileSystemResouce` class.
The `location` is never null, which is forced by the constructor of the class, so added methods don't perform any extra null checks.
```java
public FileSystemResource(String location) {
    this.location = new File(location);
}
```
Please note that [ClassPathResource](https://github.com/flyway/flyway/blob/master/flyway-core/src/main/java/org/flywaydb/core/internal/util/scanner/classpath/ClassPathResource.java#L114-L130) analogue implements equals & hashCode.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/util/scanner/filesystem/FileSystemResource.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/util/scanner/filesystem/FileSystemResource.java
@@ -104,4 +104,17 @@ public class FileSystemResource implements Resource, Comparable<FileSystemResour
     public int compareTo(FileSystemResource o) {
         return location.compareTo(o.location);
     }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) return true;
+        if (!(obj instanceof FileSystemResource)) return false;
+        final FileSystemResource that = (FileSystemResource)obj;
+        return location.equals(that.location);
+    }
+
+    @Override
+    public int hashCode() {
+        return location.hashCode();
+    }
 }


axelfontaine:
Thanks. What specific problem does this solve?

evoevodin:
I have a class holding the resource which can be either `ClassPathResource`, or `FileSystemResource` or `null`.

```java
class MigrationScript {
   private final Resource resource;
   // .. the other fields 
}
```

This class implements `hashCode` and `equals` in a classic manner:

```java
    @Override
    public boolean equals(Object obj) {
        if (this == obj) return true;
        if (!(obj instanceof MigrationScript)) return false;
        final MigrationScript that = (MigrationScript)obj;
        return Objects.equals(resouce, that.resource()) // && the other fields.
    }

    @Override
    public int hashCode() {
        int hash = 7;
        hash = 31 * hash + Objects.hashCode(resource);
        // the other fields
        return hash;
    }
```
Everything worked fine with `ClassPathResource` but i started getting NPEs when MigrationScript instance was created  with `FileSystemResource`, so i added all the necessary checks to my code
and now it uses `resource.getLocation()` directly, but:
1 - i wouldn't say it is convenient to use it in such a way
2 - i think that if an implementation of the `Resource` provides it's own equals & hashCode then 
it's mostly preferable to reuse those methods instead of picking values exposed by instance(like with `resource.getLocation()`). 
3 - when i see that `ClassPathResource` implements `equals` & `hashCode` i expecte the same from the other  `FileSystemResource` core implementation.

PR #1469, opened 2016-11-17
By default, deactivate the CommercialDBTest profile. It can always be

jochenw:
activated, but deactivating a profile is not so easy. (If possible at
all.)


axelfontaine:
Deactivating a profile is actually very straightforward. See https://github.com/flyway/flyway/blob/master/.travis.yml#L9

PR #1470, opened 2016-11-18
Add support for Oracle editionable objects

t3hc13h:
and relax the whitespace requirement when checking for delimiter change.


axelfontaine:
Thanks Andrew. Please be more careful and _run the tests_ in the future as your change contained a couple of bugs and actually broke a number of things.

Issue #1474, opened 2016-11-22
ClassCastException in AndroidScanner


benoitdevos:
Flyway sometimes crashes on Android 4.2 - 4.4, because a wrong ClassLoader is injected in AndroidScanner.
`java.lang.ClassCastException: android.app.LoadedApk$WarningContextClassLoader cannot be cast to dalvik.system.PathClassLoader`
It is not clear why Android sometimes decides to use this WarningContextClassLoader, but it's a valid ClassLoader and should then be supported by Flyway.

###### What version of Flyway are you using?
4.0.2

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin, SBT plugin, ANT tasks)
Java API

###### What database are you using (type & version)?
SQLiteDatabase 
(android.database.sqlite.SQLiteDatabase, api level 24 )

###### What operating system are you using?
Android 4.2 (JellyBean), Android 4.4 (KitKat)

###### What did you do?
(Please include the content causing the issue, any relevant configuration settings, and the command you ran)
```
        DroidDataSource dataSource = new DroidDataSource(context.getPackageName() + "/databases", dbName);
        ContextHolder.setContext(context);
        flyway = new Flyway();
        flyway.setDataSource(dataSource);
        flyway.setBaselineVersionAsString("1.0");
        flyway.setBaselineOnMigrate(true);
        flyway.setBaselineDescription("Initial baseline");
        flyway.info();
```

where context is the Application Context of my Android App.

###### What did you see instead?
An exception is thrown : 
```
java.lang.ClassCastException: android.app.LoadedApk$WarningContextClassLoader cannot be cast to dalvik.system.PathClassLoader
    org.flywaydb.core.internal.util.scanner.classpath.android.AndroidScanner.<init> (AndroidScanner.java:46)
    org.flywaydb.core.internal.util.scanner.Scanner.<init> (Scanner.java:38)
    org.flywaydb.core.Flyway.execute (Flyway.java:1353)
    org.flywaydb.core.Flyway.info (Flyway.java:1040)
```

benoitdevos:
A [pull request](https://github.com/flyway/flyway/pull/1475) has been submitted.

axelfontaine:
PR merged. Thanks!

PR #1475, opened 2016-11-22
Fixed ClassCastException in AndroidScanner #1474

benoitdevos:
Local classLoader typed to ClassLoader, not PathClassLoader
Removed unnecessary cast causing error.


axelfontaine:
Thanks for investigating and fixing this! Truly appreciated!

Issue #1364, opened 2016-07-13
Validate with target=current should fail on unapplied migration


PeeZu:
###### What version of Flyway are you using?

4.1 via ANT
###### What database are you using (type & version)?

Oracle 11.2.0.4
###### What operating system are you using?

Window 7
###### What did you do?

(Please include the content causing the issue, any relevant configuration settings, and the command you ran)

When we try to validate with property flyway.target=current, Flyway doesn't detect not applied migration.

Step to reproduce:
1. Create migrations files: V1.0_foo.sql + V2.0_bar.sql
2. run 'migrate' command
3. Create migration file: V1.1_foo.sql
4. run 'validate' command => Validate failed: Detected resolved migration not applied to database: 1.1 => Expected
5. run 'validate'  command with flyway.target=current
###### What did you expect to see?

Same as step 4 above: Validate failed: Detected resolved migration not applied to database: 1.1 
###### What did you see instead?

 Successfully validated XXX migrations.

PeeZu:
Additional info:

With the above scenario, run 'info' command with flyway.target=current.

You will see that not applied migration file is consider to **>Target**

This issue is a little bit critical because it avoid to test if current version is correctly applied.

Issue #1363, opened 2016-07-13
ValidateOnMigrate should fail for unapplied migration when outOfOrder=false


PeeZu:
###### What version of Flyway are you using?

4.1  via ANT (also tested in 4.3 => same behaviour)
###### What database are you using (type & version)?

Oracle 11.2.0.4
###### What operating system are you using?

Windows 7
###### What did you do?

(Please include the content causing the issue, any relevant configuration settings, and the command you ran)

When we try to migrate with property flyway.validateOnMigrate=true, Flyway doesn't detect not applied migration:

Step to reproduce:
1. Create migrations files: V1.0_foo.sql + V2.0_bar.sql
2. run 'migrate' command
3. Create migration file: V1.1_foo.sql
4. run 'validate' command => Validate failed: Detected resolved migration not applied to database: 1.1 => Expected
5. run 'migrate' command with flyway.validateOnMigrate=true
###### What did you expect to see?

Same as step 4 above: Validate failed: Detected resolved migration not applied to database: 1.1 

Action to do:
Migrate should be run with flyway.outOfOrder=true
###### What did you see instead?

Schema "XXXX" is up to date. No migration necessary.

rainoko:
same behavior in v4.01 and v4.0.3
didnt try v4.0.2

mmuth:
We have the same issue with v4.0.3.
It's a pity that it doesn't detect the unapplied migrations, because it is one of the core features of Flyway...

jkuipers:
I just encountered this with v4.0.3 as well. This is a very serious bug, as it means that the guarantees Flyway is supposed to provide no longer hold. We ran into this when someone had forgotten to merge a migration to a release branch, so it only showed up later after newer migrations had already been applied. 
We have not enabled out-of-order, so Flyway should detect that there's an unapplied migration file that can no longer be applied, but it simply ignores the file.

I'd be happy to contribute a patch if I knew where the error should be thrown in the case of validating while migrating. Right now the code in `MigrationInfoImpl#validate` only performs the check to detect the `IGNORED` migration if `context.pending` is `false`, but it's set to `true` by `Flyway#migrate` when calling the `doValidate` method.

Should the `DbMigrate#migrate` method then somehow detect this migration? Or should the validation already catch this?

jkuipers:
BTW, this seems like a duplicate of https://github.com/flyway/flyway/issues/1364 (filed by the same user)

mmuth:
[...]

PR #1490, opened 2016-12-12
Fix for #1363 (and #1364) caused by broken refactoring in commit 88c0…

jkuipers:
…6bb74c473db5c744636e4ccee24082139eef


mmuth:
A Merge would really be useful here :)...

Issue #1398, opened 2016-08-08
Log skipped/ignored migrations


davidkarlsen:
###### What version of Flyway are you using?

4.0.3
###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin, SBT plugin, ANT tasks)

maven plugin
###### What database are you using (type & version)?

Oracle 12.1
###### What operating system are you using?

Linux RHEL 7.x
###### What did you do?

We have scripts that will be skipped/ignored due to running with outofOrder=false

eg:

4.1 <-- this was applied before
4.2 <-- this one will rightly be skipped - I want that to be logged in order to make the user aware of this
5 <-- was aldready migrated here
###### What did you expect to see?

some warning logs, telling us that "4.2 will not be applied due to outOfOrder=false, see https://flywaydb.org/documentation/faq#hot-fixes
###### What did you see instead?

nothing special about the migrations that won't be applied - making hard to detect.

axelfontaine:
Is this still relevant now that we have `ignoreIgnoredMigrations` which can be set to `false` to fail fast?

approximate:
Yes, logging anything that gets skipped or ignored is actually pretty useful, otherwise the only way to get this info is to use "-X" command-line option.

MikielAgutu:
Thanks for the feature request. We think the existing workarounds are sufficient, so we won't be implementing this.

PR #1491, opened 2016-12-13
Log ignored migrations

khatkar:
Fixing issue #1398.
Added warning log for ignored migrations.


axelfontaine:
This fix is incorrect. It will also log when just invoking info.

PR #1502, opened 2017-01-06
Issue #181

Eviradnus:
Hello,

I addressed #181 for my own use; maybe this commit will help somebody else.

I introduced a `migrateAllOrNothing` parameter to enable atomic migrate commands (disabled by default).

Basically, it (optionally) wraps all the `DbMigrate.migrate()` logic in a global transaction, while disabling per-migration transactions and throwing a rollback-inducing exception on failure.

I didn't add appropriate unit tests and documentation, but if this PR were deemed acceptable, I could take the time to do so.

HTH

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
@@ -16,6 +16,13 @@
 package org.flywaydb.core;
 
 
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.*;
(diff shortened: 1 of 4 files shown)


hmvs:
Would be cool to have it!

nhoughto:
What is missing for this to be accepted? Would be a very useful addition, otherwise we will progress with this on our own fork.

Eviradnus:
Issue closed by https://github.com/flyway/flyway/commit/053ad59be339a565a1e9ecbfaa6b9d3e0678a7d9

PR #1506, opened 2017-01-19
Throw exception if no migration found

### LINKED ISSUE #1496: Ability to detect if no migration files were found
Currently, if Flyway is configured to look for migrations in the wrong path the following error is logged:

> Dec 25, 2016 9:27:01 PM org.flywaydb.core.internal.util.scanner.classpath.ClassPathScanner getLocationUrlsForPath
WARNING: Unable to resolve location classpath:a/b/c

I'd like to be able to detect this situation from the API in order to fail the build or run that depends on such migration files being found.

Ideally I should be able to get a list of migration locations that were referenced, but not found. Another approach could be to list the migration files found (whether they have already been found or not) so that I can throw an exception if the number is zero.

khatkar:
Added property `checkLocation` for FlywayConfiguration. If this property is set to false
(default), then flyway will only log a warning if no migration script is found at a location.
If set to true, flyway will throw an exception. Unit tests for this feature are also added.

This feature is helpful if users want their program to fail fast when no migrations are detected
at a given location.

Fixes #1496

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
@@ -91,6 +91,15 @@ public class Flyway implements FlywayConfiguration {
     private Locations locations = new Locations("db/migration");
 
     /**
+     * Check that migration scripts location exists.
+     *
+     * If set to true, flyway will check the existence of migration scripts in every location, and
+     * will throw an exception if not found. If set to false, flyway will only produce warning logs
+     * notifying the user of locations that do not contain any migration scripts. (default: {@code false})
+     */
+    private boolean checkLocation;
+
+    /**
      * The encoding of Sql migrations. (default: UTF-8)
      */
     private String encoding = "UTF-8";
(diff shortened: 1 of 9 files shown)


cowwoc:
This PR fixes my specific problem but at the same time it feels kind of rigid. I think we would be better off allowing users to pull this information than assume that they want to throw an exception if individual location entries are empty. There are other use-cases where this might not be true (e.g. the location contains multiple entries, one of which only exists in test mode). 

Maybe I'm worrying about nothing (in which case, go ahead and merge this). My counter-proposal is to let users query what files (a list of paths) were found in each location. Users can then do whatever they want with that information.

khatkar:
@cowwoc I agree. At this time the PR only reflects what spring-flyway will do when `checkLocation` is set to true. Getting a list of files found by flyway would require a bit more changes to the code base. I will prioritize this if couple more users need this feature. For now, you can do flyway.info() to find a list of migrations available to flyway, although this is not a good programmable option.

cowwoc:
@khatkar `flyway.info()` won't help. If you've applied all available migrations then there is no way of differentiating between the following two scenarios:

1. Migrations applied in database, but files missing from disk.
2. Migrations applied in database, files present on disk.

`flyway.info().applied()` only talks about the state of the migrations in the database. It won't tell you whether the corresponding files are missing on disk. Perhaps this is another option for this PR (adding this information).

axelfontaine:
Closing. See my comments on #1496

Issue #865, opened 2014-10-16
Custom tablespace for schema history table


axelfontaine:
- Optional support for databases that have this feature.
- Use default tablespace unless set
- Ignored for databases that don't support this

crgardner:
Is there now first class support in Flyway to specify the tablespace in Oracle?  If so, how do you configure it?

sbelan:
@axelfontaine, I tried custom table space with latest beta version and its still failing at the time of alter table. Please find below details.
Flyway gradle version: 6.0.0-beta2

Before using beta version I was getting below exception
`ERROR: permission denied for tablespac xyz`

Now I am using beta version and provided parameter **flyway.tablespace=xyz** and getting below exception

`
Migration  failed
  -----------------
  SQL State  : 42501
   Error Code : 0
  Message    : ERROR: permission denied for tablespace xyz
   Location   :  ()
  Line       : 14
  Statement  : ALTER TABLE "postgres"."flyway_schema_history" ADD CONSTRAINT "flyway_schema_history_pk" PRIMARY KEY ("installed_rank")`

Could you please help?

PR #1510, opened 2017-01-23
Using "dbname" for tablespaces in db2zos  

adifri:
Hi 
this pull request make possible to use Flyway with db2zos because we can set the dbname which is needed for tablespaces as placeholder in conf.

I'm using it now on our projects but it would be helpful to integrate it in a future release.

Thanks


axelfontaine:
Thanks for the pull request. I'll be honest, this has zero chance of being accepted as is. The DB name setting is *way* too specific as it only applies to DB2 z/OS which is a very niche product.

As an alternative how about tackling #865 ? This would tie in nicely with this as you could set up the tablespace as you want and then simply refer to it using the Flyway property.

Closing for now. Please feel free to reopen a PR for #865 and I'll work with you to provide you feedback.

adifri:
Hi Axel, thanks for your feedback. 
how should this new Flyway property be called? 
Would "flyway.tablespace.xxxx" like the placeholders  be a good one?
Then you can use it like this: 
flyway.tablespace.name
flyway.tablespace.dbname
flyway.tablespace.xxxx

In oracle you can also define a tablespace for a table with a specific name.
 
But you have to give it as param anyway in Flyway.java on creating MetaDataTable
`MetaDataTable metaDataTable = new MetaDataTableImpl(dbSupport, schemas[0].getTable(table), installedBy, tablespace);`

then on create MetaTable you could do something like that:
```
iterateOnTablespace {
    placeholders.put(tablespaceName,tablespaceValue);
}
```

Would that have a chance?

axelfontaine:
The property should be called `flyway.tablespace` and its configuration should affect the creation of the table named with `flyway.table` in all databases that support tablespaces (Oracle, PostgreSQL, DB2 ,...)

nycjay:
In DB2 (at least LUW), you can specify 0-3 tablespaces for a table (more if partitioned, but I don't think anyone would need to partition a Flyway table). For our tables, we specify all 3 tablespaces, so

```
create table flywayversion (
columns...
) IN tablespace1 INDEX IN tablespace2 LONG IN tablespace3;
```

If Flyway is adding support for tablespaces, could we also get support for a map of tablespaces so we could specify each? Currently, we let Flyway create its table, then we run an admin utility to move it to the 3 tablespaces where we want it to live.

axelfontaine:
@nycjay No, I feel this would needlessly complicate things as this is a very niche requirement.

PR #1513, opened 2017-01-29
Retry to obtain db connection during mojo execution.

VboxNick:
Problem:
A build pipeline might include additional steps in order to fire up a database instance, e.g. a mysql instance inside docker container. The startup process might be asynchronous and Flyway Maven Plugin will crash immediately. 

Solution:
Added an optional parameter 'maxRetries'. The plugin will try to obtain connection every 1 second till connection will be obtained or 'maxRetries' reached. When 'maxRetries' is reached, an exception will be thrown. The parameter 'maxRetries' has default value 0, that means retries won't be performed.

### DIFF
--- a/flyway-maven-plugin/pom.xml
+++ b/flyway-maven-plugin/pom.xml
@@ -56,6 +56,11 @@
             <artifactId>h2</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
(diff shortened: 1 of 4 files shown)


axelfontaine:
Thanks for the pull request. This is however something that when implemented, should be part of the core and not the Maven plugin. Also it should probably be solved transparently inside a datasource.

VboxNick:
You're right. Shall I implement this? Maybe you can guide me in terms of the best place where it should be. 

Thanks for your time.

Issue #1514, opened 2017-01-30
java.lang.ClassCastException: org.flywaydb.gradle.FlywayExtension_Decorated cannot be cast to org.flywaydb.gradle.FlywayExtension


donbeave:
###### What version of Flyway are you using?

4.0.3

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin, SBT plugin, ANT tasks)

Gradle plugin

###### What database are you using (type & version)?

PostgreSQL

###### What operating system are you using?

Mac OS X

###### What did you do?
(Please include the content causing the issue, any relevant configuration settings, and the command you ran)

```groovy
        task "someTestMigration"(type: FlywayMigrateTask) {
            user = datasourceUsername
            password = datasourcePassword
            locations = flywayLocations
            url = datasourceUrl
        }
```

###### What did you expect to see?

Workable `./gradlew someTestMigration`

###### What did you see instead?

```
Caused by: org.gradle.api.tasks.TaskInstantiationException: Could not create task of type 'FlywayMigrateTask'.
        at org.gradle.api.internal.project.taskfactory.TaskFactory$1.call(TaskFactory.java:123)
        at org.gradle.api.internal.project.taskfactory.TaskFactory$1.call(TaskFactory.java:118)
        at org.gradle.util.GUtil.uncheckedCall(GUtil.java:402)
        at org.gradle.api.internal.AbstractTask.injectIntoNewInstance(AbstractTask.java:178)
        at org.gradle.api.internal.project.taskfactory.TaskFactory.create(TaskFactory.java:118)
        at org.gradle.api.internal.project.taskfactory.TaskFactory.createTask(TaskFactory.java:77)
        at org.gradle.api.internal.project.taskfactory.AnnotationProcessingTaskFactory.createTask(AnnotationProcessingTaskFactory.java:46)
        at org.gradle.api.internal.project.taskfactory.DependencyAutoWireTaskFactory.createTask(DependencyAutoWireTaskFactory.java:39)
        at org.gradle.api.internal.tasks.DefaultTaskContainer.create(DefaultTaskContainer.java:76)
        at org.gradle.api.internal.project.DefaultProject.task(DefaultProject.java:1038)
        at org.gradle.internal.metaobject.BeanDynamicObject$MetaClassAdapter.invokeMethod(BeanDynamicObject.java:382)
        at org.gradle.internal.metaobject.BeanDynamicObject.invokeMethod(BeanDynamicObject.java:170)
[...]

PR #1515, opened 2017-02-01
Fix for "FlywayExtension_Decorated cannot be cast to org.flywaydb.gradle.FlywayExtension"

donbeave:
https://github.com/flyway/flyway/issues/1514


axelfontaine:
The code seems identical, except for the setClassLoader part. How exactly does this fix the issue?

axelfontaine:
This fix actually breaks the loading of the JDBC drivers.

PR #1522, opened 2017-02-10
Update upgradeMetaDataTable.sql

anujva:
The create script doesn't have the vr_idx or ir_idx. When it runs upgrade script it fails.


axelfontaine:
Yes it does: https://github.com/flyway/flyway/blob/flyway-3.2.1/flyway-core/src/main/resources/org/flywaydb/core/internal/dbsupport/postgresql/createMetaDataTable.sql

Issue #1531, opened 2017-02-15
SAP HANA: CREATE PROCEDURE statements with embedded semicola don't work


ace130-github:
###### What version of Flyway are you using?
Current master (03136dce684447fb1d3856905fdcd3e9ff531063, date: Feb 15, 2017) 

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin, SBT plugin, ANT tasks)
Command line

###### What database are you using (type & version)?
SAP HANA Express 2.0 SPS 0

###### What operating system are you using?
Windows 7

###### What did you do?
Try to have flyway import the following file:
```
-- just to have one statement before the actual test statement
CREATE VIEW all_misters AS SELECT * FROM test_user WHERE name LIKE 'Mr.%';

-- create statement with embedded ';'
-- it is expected that this statement is read as a single statement ending after 'END;'
-- and not broken into separate statements ending after each ';' withing the BEGIN...END-block.
-- this example statement is borrowed from the HANA reference for
-- CREATE PROCEDURE (http://help-legacy.sap.com/saphelp_hanaplatform/helpdata/en/20/d467407519101484f190f545d54b24/content.htm)
CREATE PROCEDURE orchestrationProc
 LANGUAGE SQLSCRIPT AS
 BEGIN
   DECLARE v_id BIGINT;
   DECLARE v_name VARCHAR(30);
   DECLARE  v_pmnt BIGINT;
   DECLARE v_msg VARCHAR(200);
   DECLARE CURSOR c_cursor1 (p_payment BIGINT) FOR
     SELECT id, name, payment FROM control_tab
       WHERE payment > :p_payment ORDER BY id ASC;
   CALL init_proc();
   OPEN c_cursor1(250000);
   FETCH c_cursor1 INTO v_id, v_name, v_pmnt; v_msg := :v_name || ' (id ' || :v_id || ') earns ' || :v_pmnt || ' $.';
   CALL ins_msg_proc(:v_msg);
   CLOSE c_cursor1;
 END;

-- just to have a trailing statement for testing
DROP VIEW all_misters;
```
which is syntactically correct in the SAP HANA SQL dialect (in fact, the main part is copied from the HANA SQL reference for CREATE PROCEDURE).

###### What did you expect to see?
The file should be split into three statements (CREATE VIEW..., CREATE PROCEDURE..., DROP VIEW) and the statements should be executed against the DB.

###### What did you see instead?
[...]

PR #1532, opened 2017-02-15
#1531: Fix for CREATE PROCEDURE in SAP HANA dialect

ace130-github:
This is a preliminary fix for #1531.

Some things might  be reconsidered:

- I'm not sure that the test and its associated resource are in the correct place
- I wonder if it's appropriate to just throw a RuntimeException in case the syntax expectations are not met. Actually, I can't think of a more appropriate way to handle this situation. In any case, there might be a better suited subclass of RuntimeException.
- There are some assumptions (ie BEGIN at the beginning, END; at the end of a line) that could possibly be at least documented.
- And finally, there are a couple of other CREATE x-Statements (CREATE FUNCTION, ...) that might also benefit from that change, they could just be added to the if...-Statement.

Please let me know if the fix is going in the right direction, then I can work a little more on these items.


flyway-core/src/test/java/org/flywaydb/core/internal/dbsupport/saphana/SapHanaSqlStatementBuilderSmallTest.java
```diff
@@ -77,4 +83,54 @@ public void temporalString() {
...
+                statements.add(statementBuilder.getSqlStatement().getSql());
+                statementBuilder = new SapHanaSqlStatementBuilder();
+            }
+        }
+        assertEquals("Should recognize exactly three statements", 3, statements.size());
+    }
+
+    private static final String PROCEDURE_TEST_SCRIPT_LOCATION = "/org/flywaydb/core/internal/dbsupport/saphana/procedures/procedure.sql";
+
+    private static String readResource(String name) {
```
axelfontaine: This can be replaced with new ClassPathResource(...).loadAsString("UTF-8")

flyway-core/src/main/java/org/flywaydb/core/internal/dbsupport/saphana/SapHanaSqlStatementBuilder.java
```diff
@@ -44,22 +40,32 @@ protected String cleanToken(String token) {
 
     @Override
     protected Delimiter changeDelimiterIfNecessary(String line, Delimiter delimiter) {
-        if (StringUtils.countOccurrencesOf(statementStart, " ") < 4) {
```
axelfontaine: This change actually breaks the case where CREATE and PROCEDURE are on separate lines

flyway-core/src/main/java/org/flywaydb/core/internal/dbsupport/saphana/SapHanaSqlStatementBuilder.java
```diff
@@ -44,22 +40,32 @@ protected String cleanToken(String token) {
...
-            statementStart += line;
-            statementStart += " ";
+
+        // TODO: When a line contains a CREATE X and and END statement, then the order is significant.
+        // TODO: There are a couple more statements allowing for nested begin/end blocks
+        if (line.startsWith("CREATE PROCEDURE") || line.startsWith("CREATE TRIGGER")) {
```
axelfontaine: Why is this nesting necessary? Can you point to a valid example SQL or better: add a test?

flyway-core/src/test/java/org/flywaydb/core/internal/dbsupport/saphana/SapHanaSqlStatementBuilderSmallTest.java
```diff
@@ -77,4 +83,54 @@ public void temporalString() {
...
+            statementBuilder.addLine(line);
+            if (statementBuilder.isTerminated()) {
+                statements.add(statementBuilder.getSqlStatement().getSql());
+                statementBuilder = new SapHanaSqlStatementBuilder();
+            }
+        }
+        assertEquals("Should recognize exactly three statements", 3, statements.size());
+    }
+
+    private static final String PROCEDURE_TEST_SCRIPT_LOCATION = "/org/flywaydb/core/internal/dbsupport/saphana/procedures/procedure.sql";
```
axelfontaine: Just inline this.

ace130-github:
Thanks for your valuable feedback, I'll take some time to look into this.

flyway-core/src/test/resources/org/flywaydb/core/internal/dbsupport/saphana/procedures/procedureWithNestedBlocks.sql
```diff
@@ -0,0 +1,23 @@
+-- just to have one statement before the actual test statement
+CREATE VIEW all_misters AS SELECT * FROM test_user WHERE name LIKE 'Mr.%';
+
+-- a procedure with nested begin-end; blocks
+-- also taken from
+-- SAP HANA SQLScript Reference - SAP Library at:
+-- http://help-legacy.sap.com/saphelp_hanaplatform/helpdata/en/f0/a6dceac8b94cca98dd2741ac6541b8/content.htm?frameset=/en/c6/558a64245942ebb52f6a6ddb3e4278/frameset.htm
```
axelfontaine: I can't see an example with nested blocks there. Where did you see it?

flyway-core/src/main/java/org/flywaydb/core/internal/dbsupport/saphana/SapHanaSqlStatementBuilder.java
ace130-github: Yes, BEGIN...END; blocks can be nested.
I have attached a test for that case.

flyway-core/src/main/java/org/flywaydb/core/internal/dbsupport/saphana/SapHanaSqlStatementBuilder.java
ace130-github: I found the approach useful to determine the start of the statement we're building. But it only worked in very limited cases. I tried to extend it by ignoring multiple blank lines, line comments and the beginning of block comments.

I also added tests specifically for the case you mentioned.

flyway-core/src/test/java/org/flywaydb/core/internal/dbsupport/saphana/SapHanaSqlStatementBuilderSmallTest.java
ace130-github: OK

flyway-core/src/test/java/org/flywaydb/core/internal/dbsupport/saphana/SapHanaSqlStatementBuilderSmallTest.java
ace130-github: OK

flyway-core/src/test/resources/org/flywaydb/core/internal/dbsupport/saphana/procedures/procedureWithNestedBlocks.sql
ace130-github: Sorry, I used the "share" link on the SAP Help page, which seems not to include the full link.
I updated the link in file.

ace130-github:
I resolved all your comments and added a couple of tests for these scenarios. In addition, I'll leave some replies to your comments explaining what I did.

I can still think of a thousand ways to break this functionality:
- Have BEGINs not at the beginning and END;s not the end of a line, or both together in one line
- CREATE PROCEDURE statements in "inner" lines of block comments
- ...

In order to solve all these issues properly, we would have to have some kind of parser that's fully aware of the syntax (we're already implementing a part of it in this functionality). But this would be a major change in the base class, and would be misplaced to implement just in this subclass.

So I think we have covered fairly common cases which is better than nothing.

flyway-core/src/main/java/org/flywaydb/core/internal/dbsupport/saphana/SapHanaSqlStatementBuilder.java
```diff
@@ -44,24 +41,56 @@ protected String cleanToken(String token) {
...
             if (line.startsWith("BEGIN")) {
-                insideBeginEndBlock = true;
+                beginEndNestedDepth++;
             }
 
             if (line.endsWith("END;")) {
-                insideBeginEndBlock = false;
+                beginEndNestedDepth--;
+                if (beginEndNestedDepth < 0) {
+                    throw new FlywayException("Syntax error in SQL script: unpaired END; statement");
```
axelfontaine: I wouldn't worry so much about this. A warning is probably enough here. If the script is incorrect (in itself, not due to our parsing) the DB will tell the user.

ace130-github: OK, I replaced the exception with a warning in the logs,
Glad to see the pull requests were added to the upcoming milestone.

Issue #1550, opened 2017-03-12
Numerous flaws in Clean implementation for Oracle DB


vosolovskiy:
#####  Flaws in Clean phase:
- [x] Protection from cleaning SYSTEM schema should be extended to all Oracle-maintained schemas. The list of schemas can be found as `select schema, other_schemas from dba_registry`.
- [x] Flag `defaultSchemaForUser` is computed wrongly: it shouldn't ignore case when compares the schema name with the deploy user name, because schema names are always quoted in the Flyway code (which is correct). 
- [x] Implementation of `OracleDbSupport#doCurrentSchemaName` method is wrong: the correct SQL should be `SELECT SYS_CONTEXT('USERENV', 'CURRENT_SCHEMA') FROM DUAL`. New methods should be added to `DbSupport` class and its inheritors: `getCurrentUser` and `doGetCurrentUser`, which by default can return `jdbcTemplate.getMetaData().getUserName()` (which is not precise) and should be overridden. For Oracle it's `SELECT USER FROM DUAL`. This new method must be used for computing `defaultSchemaForUser` flag named above instead of `doCurrentSchemaName`.
- [x] The warning message with object types that cannot be cleaned due to Oracle limitations should be revised. AQ and Scheduler objects (and many many others) can be dropped if the deploy user has corresponding privileges. The only problems I know are related to Spatial metadata, DB links and old style Jobs. These objects still can be dropped not by the owner with a temporary local procedure or some SYS-owned packages, but this should be done in a callback, I suppose.
- [x] Implementation of `OracleDbSupport#doAllTables` to fix #178 is fallible. It should be corrected. There's no use to count references. Instead this graph should be traversed in correct direction.
- [ ] Using ALL_ data dictionary views. According to [this article](https://mwidlake.wordpress.com/2009/09/21/) it is recommended to use DBA_ views when possible for better performance. Availability of DBA_ views can be checked before running Clean command. Superusers allowed to drop/create others' objects usually have access to DBA_ views.
- [ ] The body of `doClean()` method should be refactored (and this remark is related to all the `Schema` implementations): there are a lot code duplication (repeated typical for-loops) and violation of SRP (clean logic for each and every object type/metadata is stored in the same class). This makes `doClean()` less supportable. 
[...]

PR #1552, opened 2017-03-12
Fixed the current schema/current user confusion in OracleSchema.

vosolovskiy:
This pull request is to address some of the flaws mentioned in Issue #1550 related to the confusion between current_schema and user functions in Oracle database.

Key points to keep in mind:
1. Session user and current schema may differ.
2. JDBC connection user (jdbc metadata) and Oracle session user (select user from dual) may differ.
3. Oracle session user cannot modify or needs extra privileges to modify objects in schemas different from its own schema.
4. User/schema names are case-sensitive. Flyway uses quoted schema names to work correctly.


flyway-core/src/main/java/org/flywaydb/core/internal/dbsupport/DbSupport.java
```diff
@@ -144,6 +143,30 @@ public void restoreCurrentSchema() {
     protected abstract void doChangeCurrentSchemaTo(String schema) throws SQLException;
 
     /**
+     * Retrieves the current user.
+     *
+     * @return The current user for this connection.
+     */
+    public String getCurrentUserName() {
```
axelfontaine: What is the use of this method (except for the Oracle-specific test you added)?

axelfontaine:
Thanks for the pull request. Looks very good. See my comment above for a small clarification.

flyway-core/src/main/java/org/flywaydb/core/internal/dbsupport/DbSupport.java
vosolovskiy: User and Schema are different terms in most databases. This method should be the only correct way of determining the current user. `jdbcTemplate.getMetaData().getUser()` may lie (and it does for proxy connections). That's why this method should be considered for overriding in database-specific subclasses, and is overriden in OracleDbSupport class.

Absence of this method may lead (and actually did lead) to misuse of another method - `getCurrentSchemaName`, which was wrongly implemented (intentionally?) and works correctly in wrong context and incorrectly in right context.

This method is now used in doClean method instead of misused getCurrentSchemaName method.

vosolovskiy: If you meant the public method `getCurrentUserName`, not its protected implementation method `doGetCurrentUserName`, then the answer to you question is - this public method should be a part of Flyway API, just as all other public methods of this class. There may be different use cases for it.

axelfontaine: My point is that this is probably just an Oracle-specific thing. Since Oracle is really the only database blurring the line between schema and user and therefore it probably doesn't belong in this general interface and should be kept to the Oracle-specific part.

vosolovskiy: Many databases differentiate Session User, Current User (user name of current execution context) and Current Schema. Among them are Oracle, PostgreSQL, MS SQL Server and maybe others. This is a part of the SQL standard, so it may be useful for any database.

But I agree that putting this method to DbSupport class with a default implementation via JDBC MetaData may be erroneus at this moment. Let me keep it in OracleDbSupport and let's consider adding these methods in future, when implementations for all other databases are ready and tested.

Will fix it today.

vosolovskiy: Fixed. Please review.

vosolovskiy:
Cancel due to a wrong branch selection. GitHub still doesn't support changing the fork branch in a pull request.
Superseded by #1558

Issue #1549, opened 2017-03-10
Flyway "(unsupported protocol: war)" with Tomcat 8.0.41 and unpackWARs=false


thomasmey:
Hi,

this Tomcat commit https://github.com/apache/tomcat80/commit/7e767cc6efe79cdd367213da3c1f88711a29ad7a#diff-a72fb99b0729353084d2c437f749e718 broke Flyway in https://github.com/flyway/flyway/blob/7925112a1041d7763d722a4d36cbbb39277e7b5c/flyway-core/src/main/java/org/flywaydb/core/internal/util/scanner/classpath/ClassPathScanner.java#L310
with Tomcat config unpackWARs=false.
The returned URL is now war: instead of jar:

thomasmey:
The fix from https://github.com/flyway/flyway/issues/1512 is incomplete as the JarFileClassPathLocationScanner will get used for protocol "war" but the protocol separator is then "*/" and not "!/", so getJarFromUrl() will fail to correctly determine the file name.

This whole code fragment looks very similar to https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/core/io/support/PathMatchingResourcePatternResolver.java#L578 which has a similar bug reported in https://jira.spring.io/browse/SPR-15332

axelfontaine:
Arrgghh!! OK, we'll have to adjust for this.

axelfontaine:
Thanks for the PR! Merged.

PR #1555, opened 2017-03-14
Fix Tomcat URL handling fallout

thomasmey:
Issue https://github.com/flyway/flyway/issues/1549
Untested proposed fix.
Patch against 4.1.2

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/util/scanner/classpath/ClassPathScanner.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/util/scanner/classpath/ClassPathScanner.java
@@ -308,16 +308,15 @@ public class ClassPathScanner implements ResourceAndClassScanner {
         }
 
         if ("jar".equals(protocol)
-                || "war".equals(protocol)
+                || "war".equals(protocol) //Tomcat
                 || "zip".equals(protocol) //WebLogic
                 || "wsjar".equals(protocol) //WebSphere
-                || "war".equals(protocol) //Tomcat
                 ) {
             JarFileClassPathLocationScanner locationScanner;
             if ("war".equals(protocol)) {
-                locationScanner = new JarFileClassPathLocationScanner();
-            } else {
                 locationScanner = new JarFileClassPathLocationScanner("*/");
+            } else {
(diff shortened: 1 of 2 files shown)


flyway-core/src/main/java/org/flywaydb/core/internal/util/scanner/classpath/ClassPathScanner.java
```diff
@@ -308,11 +308,16 @@ private ClassPathLocationScanner createLocationScanner(String protocol) {
...
         if ("jar".equals(protocol)
-                || "war".equals(protocol)
+                || "war".equals(protocol) //Tomcat
                 || "zip".equals(protocol) //WebLogic
                 || "wsjar".equals(protocol) //WebSphere
                 ) {
-            JarFileClassPathLocationScanner locationScanner = new JarFileClassPathLocationScanner();
+            JarFileClassPathLocationScanner locationScanner;
+            if ("war".equals(protocol)) {
+                locationScanner = new JarFileClassPathLocationScanner("*/");
```
axelfontaine: Won't this be breaking older Tomcat releases?

thomasmey: I don't think so, as older tomcat version will return URLs like jar:xxx!/ for war files, and new tomcats will return URLs like war:xxx*/. but somebody should test this...

thomasmey: Hi, I manually did test this patch against Tomcat 8.5.5 and 8.5.13 (old and new URL handling) with plain 4.1.2 and patched 4.1.2.
The patched 4.1.2 always works correctly in all four constellations.
Do you need anything else to merge this PR?

axelfontaine: OK, works for me. Thanks for verifying!

axelfontaine:
@thomasmey Could you please get in touch with me by email? (axel at boxfuse.com) Thanks!

PR #1558, opened 2017-03-18
Fixed the current schema/current user confusion in OracleSchema

vosolovskiy:
This pull request is to address some of the flaws mentioned in Issue #1550 related to the confusion between current_schema and user functions in Oracle database.

Main points:
1. Current schema name can be determined with `SYS_CONTEXT('USERENV', 'CURRENT_SCHEMA')` function.
2. Current execution context user name can be determined with `USER` function.
3. Modifying objects in shemas other than the user's own schema may require extra privileges or be impossible. Flyway checks if the current user and the affected schema coincide while executing `Clean` command.
4. User/schema names are case-sensitive and must be compared with `equals()` method. Flyway uses quoted schema names to work correctly.

This PR supersedes PR #1552 (closed due to a wrong branch selection).


axelfontaine:
Merged. Thanks for the excellent PR!

PR #1185, opened 2016-01-26
Followup to #1183: Subclassable resolvers

pauxus:
This is a followup PR to #1183. It makes the built-in resolvers more override friendly.

PR #1566, opened 2017-03-22
Make default resolvers handle all locations in one instance

pauxus:
... instead of the previous 'one resolver per single location'. This makes the default resolvers behave more like custom resolvers and is a step on the way to eventually allow subclasses of default resolvers as custom resolvers.

This PR has no direct effect in itself, but provides only the base. 

This is basically a revisit of #1183 and #1185 

@axelfontaine : Since there will be more than one commit to this topic, I would rather put them all into one PR but as separate commits. Or do you rather want single PRs for all steps?


axelfontaine:
Single PRs are easier to digest/accept/reject. I highly encourage you to write down your plan for each PR in an issue first to avoid unnecessary work if it has no chance of being accepted.

axelfontaine:
Good idea! Aligning internal resolvers with custom ones makes sense!

Thanks for the pull request!

Issue #1567, opened 2017-03-23
Subclassable resolvers


pauxus:
Currently, we have an option to skipDefaultResolvers, allowing us to completely replace the used resolvers. However, most of the time, when replacing the default resolvers, we actually want only to tweek the behaviour of them (for example: use a different strategy for caculating the versions, such as encoding part of the version in the directory name) (see also #1183 and #1185)

This is currently rather ugly, because we need to copy and paste a lot of code from the default resolvers.

I propose a couple of changes in order to make this easier. The main goal of these changes is to allow to subclass the default resolvers in order to prevent copy and paste (of course, since the default resolvers are not official API, such custom subclasses need to be tied to specific flyway versions to be guaranteed to work):

- [x] #1566 This is a first step. When creating a custom resolver, only one instance is created, while for the default resolvers, one instance is created for each location. I moved the locations loop inside the resolvers. This does not change the behaviour, but is a necessary step to allow to replace a default resolver with a custom subclass
- [x] Let the default resolvers use an empty constructor. For this, three steps are needed: (partially in #1570)
  - [x] let them take the data needed from the flywayConfiguration (#811) object injected into them. (previously implement in [5b7bfd9]) (#1569)
  - [x] We need a different way to inject the Scanner into the default resolvers. This could work via a separate setter or some other mechanism (#1568)
  - [x] Remove usage of `DbSupport` in `SqlMigrationResolver` altogether. The Resolver and Executor do not really need it, they simple hand it down to SqlScript, which could always create a matching DbSupport object using the jdbcTemplate's connection during `execute()`. This would result in a lot more DbSupport objects being created (which are cheap, but could be - if necessary - negated by caching the `DbSupport` objects with their connection as `WeakReference` in `WeakHashMap`). Also, now the parsing of the scripts would be done during the actual execution of the migration, not at resolving time (this could also be prevented by giving`SqlScript`access to the`FlywayConfiguration` object)
- [ ] documentation update and testcase. These really only make sense for the issue as a whole, since most changes are internal only and thus covered by the existing tests. However, once the issue is resolved, all a test case and documentation for the feature (replacing a resolver with a custom subclass) need to be provided.

Two more tasks which would make this even more useful (just to show where I am headed with this):
- Provide some methods of the default resolvers as protected, to allow overriding specific methods (Usecase for us: Parse the migration version not only from the filename, but also from the folder structure : 1.4_new_Feature/5_procedures/V1.2__demo_procedure --> 1.4.5.1.2)
[...]

PR #1568, opened 2017-03-23
Scanners are cached using WeakReferences

pauxus:
This allows Resolvers to have access to existing scanners without the need to pass them around. This is a preliminary step to allow full replacement of default resolvers. This effectively rolls back a previous optimization without sacrificing the cache effect.

(note that this PR is based on #1566, before that base PR is merged, the changes shown will be from both PRs)

This is part of #1567


pauxus:
Rebased

Please consider merging #1582 insteads, which groups all relevant PRs

axelfontaine:
Closing old unmerged PR as part of repo cleanup.

PR #1570, opened 2017-03-23
Default constructors for jdbc resolvers

pauxus:
(needs PR #1569)

Part of #1567 

I split this one from the handling of SqlMigrationResolver, because this one is rather trivial, while SqlMigrationResolver changes need handling for DbSupport)

PR #1569, opened 2017-03-23
Read all necessary information from FlywayConfiguration object

pauxus:
Since with the exception of DbSupport, all necessary data is already included in that 

Part of #1567 

Note that this PR depends on #1568. Before that PR is merged, this PR will include all changes from that PR as well.

PR #1581, opened 2017-03-31
Default constructor for sql resolver

pauxus:
part of implementation of #1567.

Similar to #1570. This PR changes the `SqlMigrationResolver` to also use a default constructor and injection of FlywayConfiguration. However, since it depended on an instance of `DbSupport` this is somewhat bigger:

- Take all necessary information from FlywayConfiguration
- inject Flyway configuration via ConfigurationAware
- Don't pass down DbSupport (it is only ever needed in SqlScript)
- SqlScript obtains its DbSupport via DbFactory (cheap since the connection is reused)
- Don't pass down placeholders, they can be created from FlywayConfiguration

This PR includes #1570 as well as #1569 and #1568 

I would suggest merging them all together, which would keep the history cleaner (since they are all related)


axelfontaine:
Closing old unmerged PR as part of repo cleanup.

PR #1601, opened 2017-04-11
Clean improvements for Oracle pt.2

vosolovskiy:
This PR is to address the rest of flaws mentioned in issue #1550. This PR continues my previous PR #1573.

The main goal of this change is to simplify `OracleSchema.doClean()` method and extend it with more object types. Logic of finding and dropping of objects of certain types was encapsulated into corresponding `OracleSchema.ObjectType` enum values.

The list of new object types that get dropped during Clean phase:

- DIMENSION
- INDEXTYPE
- OPERATOR
- CONTEXT
- LIBRARY
- JAVA CLASS
- JAVA RESOURCE
- PROGRAM
- SCHEDULE
- CHAIN
- FILE WATCHER
- EVALUATION CONTEXT
- RULE
- RULE SET
- FILE GROUP
- XML SCHEMA
- MINING MODEL
- REWRITE EQUIVALENCE
- SQL TRANSLATION PROFILE

There are also object types that are intentionally ignored during Clean phase, because they contain sensitive information (passwords for external resources) and not always can be re-created via standard Flyway workflow:
- DATABASE LINK
- CREDENTIAL

Other ignored schema-owned types (tricky or undocumented):

- DESTINATION
- SCHEDULER GROUP
- CUBE
- CUBE DIMENSION
- CUBE BUILD PROCESS
- MEASURE FOLDER
- ASSEMBLY (undocumented)
- JAVA DATA (undocumented)

Unlike the previous implementation, which silently ignored most of unsupported types, in the new version all the ignored (non-droppable) types cause warnings in the log if found in the target schema and still can be explicitly dropped with an afterClean-callback. Even if not cleaned, they _don't break_ migration from scratch after full cleaning - Flyway sees them during Clean phase but doesn't see them during `isEmpty()` check and doesn't throw an exception.

Medium tests were improved to support different Oracle environments. JUnit assumptions on Oracle version and edition were added where needed, so that some tests are automatically ignored if the target database doesn't support the tested feature. There's no need to set/unset `@Ignore` annotation manually any more.

This change has been successfully tested on the following Oracle versions/editions:

- 10.2.0.1 XE
- 10.2.0.5 SE
- 10.2.0.5 EE
- 11.2.0.2 XE
- 11.2.0.4 SE
- 11.2.0.4 EE
- 12.1.0.2 EE (non-PDB).


vosolovskiy:
Any chance to see it in 4.2.0? How can I assist in promoting it?

axelfontaine:
@vosolovskiy Can you get in touch with me via email? (axel at boxfuse.com) Thanks!

axelfontaine:
@vosolovskiy Thanks. Merged!I had to disable one test for XE. As you may have noticed we have started transitioning Flyway testing to Docker containers to make it easier for contributors to have consistent dev environments. One test (`scheduler112Enhancement`) however does not appear to work with the Oracle XE 11.2 container we use. I assume it probably requires SE as it failed with `ORA-03001: unimplemented feature`, so that's how I modified it. Please correct me if I'm wrong.

vosolovskiy:
@axelfontaine cool, thanx!

PR #1607, opened 2017-04-16
For Postgresql migrations, add support for "DROP INDEX CONCURRENTLY .*",

jpswain:
which must run outside of a transaction.
https://www.postgresql.org/docs/9.6/static/sql-dropindex.html


axelfontaine:
Good catch! Thanks!

PR #1611, opened 2017-04-24
Fix a minor typo in Javadoc

durimkryeziu:
While reading the FlywayCallback docs faced that typo and changed it from **will** to **with**


axelfontaine:
Good catch! Thanks!

Issue #1631, opened 2017-05-10
Flyway Gradle plugin should not add JAR(s) from a project's source sets for scanning if resolved artifact file does actually not exist.


twwwt:
###### What version of Flyway are you using?
4.2.0

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin, SBT plugin, ANT tasks)
Gradle 3.5.0

###### What database are you using (type & version)?
irrelevant to this issue

###### What operating system are you using?
irrelevant to this issue

###### What did you do?
In a multi-project Gradle build, consider a sub project, say *s_1*, that applies the Gradle Flyway plugin. In addition, assume that *s_1* defines a project dependency onto another sub project, say *s_2*. If the `locations` property contains at least one `classpath:` reference, the Flyway plugin will add resolved artifacts from the source sets specified by *s_2* to the Java-based migrations scanner (besides those of *s_1*). This is the case even if there are actually no Java-based migrations in any of the source sets defined by *s_2*. Next, run any of the Flyway Gradle tasks on *s_1*; e.g.

```gradle :s_1:flywayInfo```

Note that in this example, it was intended to run `flywayInfo` **just** for *s_1*. As long as `flywayInfo` does not depend on some task of *s_2* (e.g. the `build` task), no task of *s_2* will run. If *s_2* is "clean" at this point, no artifacts of source sets specified by *s_2* do exist; but `flywayInfo` will still add them for scanning (since Gradle resolves them at configuration phase).

One possible solution to this behavior is to let `:s_1:flywayInfo` depend on `:s_2:build`. That, however, would be a rather poor solution since building *s_2* in order to be able to execute `:s_1:flywayInfo` is not what one wants; in particular, if it is known that *s_2* does **not** contribute any Java-based migrations.

###### What did you expect to see?
The build of *s_1* should complete without errors. Considering the example, it should print information about the database migration state (taking into account migrations found according to how locations have been configured).

###### What did you see instead?
The build ends by throwing an exception (private informations have been removed/modified):
```
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':s_1:flywayInfo'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeActions(ExecuteActionsTaskExecuter.java:98)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:68)
[...]

PR #1632, opened 2017-05-10
Skip artifacts from adding them to the scanner that do not exist.

twwwt:
Proposed solution to issue #1631


twwwt:
Rethinking that pull request, I think it should **not** be applied. The reason is that the behavior of skipping an artifact that does not exist should take place as late as possible; that is, the moment when the scanner actually runs. The pull request implements eager behavior: an artifact is not added if it does not exist at Gradle's configuration phase. In general, an artifact may very well be created only after configuration phase.

In summary, the additional test whether a file/artifact does actually exist should be implemented directly inside Flyway's scanner, not in the Gradle plugin.

axelfontaine:
@twwwt I hadn't looked at the code yet. Agreed.

Issue #1649, opened 2017-05-27
flyway:clean fails on db2 when trying to drop versioning on tables which is accessed by a MQT


msallstr:
###### What version of Flyway are you using?

4.2.0

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin, SBT plugin, ANT tasks)

maven plugin

###### What database are you using (type & version)?

db2 10.5

###### What operating system are you using?

windows/linux

###### What did you do?
(Please include the content causing the issue, any relevant configuration settings, and the command you ran)

flyway:clean

###### What did you expect to see?

[INFO] Successfully dropped schema "TESTSCHEMA" (execution time 00:10.829s)
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 17.108 s
[INFO] Finished at: 2017-05-27T14:49:59+02:00
[INFO] Final Memory: 28M/407M
[INFO] ------------------------------------------------------------------------

###### What did you see instead?

[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 7.076 s
[INFO] Finished at: 2017-05-27T14:54:01+02:00
[INFO] Final Memory: 27M/406M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.flywaydb:flyway-maven-plugin:4.2.0:clean (default-cli) on project uppfoljning: org.flywaydb.core.internal.dbsupport.FlywaySqlException:
[ERROR] Unable to clean schema "TESTSCHEMA"
[ERROR] ------------------------------------
[ERROR] SQL State  : 42997
[ERROR] Error Code : -270
[ERROR] Message    : DB2 SQL Error: SQLCODE=-270, SQLSTATE=42997, SQLERRMC=21, DRIVER=4.19.66
...
```
SQL State  : 42997
Error Code : -270
Message    : DB2 SQL Error: SQLCODE=-270, SQLSTATE=42997, SQLERRMC=21, DRIVER=4.19.66

        at org.flywaydb.maven.AbstractFlywayMojo.execute(AbstractFlywayMojo.java:602)
        at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:134)
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:207)
        ... 20 more
[...]

PR #1650, opened 2017-05-27
#1649 flyway:clean fails on db2 when trying to drop versioning on tab…

msallstr:
Make an explicit drop of all MQTs if there are tables with versioning. 

Modified the MQT test case to involve a table with versioning


axelfontaine:
@msallstr Can you sign the CLA so this can be merged? Thanks!

msallstr:
@axelfontaine done!

axelfontaine:
Thanks! Merged.

Issue #1305, opened 2016-05-06
Maven warning when using Flyway plugin with parallel builds


cowwoc:
###### What version of Flyway are you using?

4.0
###### What database are you using (type & version)?

Doesn't matter
###### What operating system are you using?

Windows 7
###### What did you do?

(Please include the content causing the issue, any relevant configuration settings, and the command you ran)

```
            <plugin>
                <groupId>org.flywaydb</groupId>
                <artifactId>flyway-maven-plugin</artifactId>
                <version>4.0</version>
                <executions>
                    <execution>
                        <id>migrate-db</id>
                        <goals>
                            <goal>migrate</goal>
                        </goals>
                        <phase>generate-sources</phase>
                    </execution>
                </executions>
                <dependencies>
                    <dependency>
                        <groupId>mysql</groupId>
                        <artifactId>mysql-connector-java</artifactId>
                        <version>${mysql.version}</version>
                    </dependency>
                </dependencies>
            </plugin>
```

Invoked "mvn clean install -T 8" to trigger parallel builds.
###### What did you expect to see?

According to https://flywaydb.org/documentation/faq#parallel Flyway supports parallel migrations, so I am not expecting to see any warnings.
###### What did you see instead?

```
[WARNING] *****************************************************************
[WARNING] * Your build is requesting parallel execution, but project      *
[WARNING] * contains the following plugin(s) that are not marked as       *
[WARNING] * @threadSafe to support parallel building.                     *
[WARNING] * While this /may/ work fine, please look for plugin updates    *
[WARNING] * and/or request plugins be made thread-safe.                   *
[WARNING] * If reporting an issue, report it against the plugin in        *
[WARNING] * question, not against maven-core                              *
[WARNING] *****************************************************************
[WARNING] The following plugins are not marked @threadSafe in data-access:
[WARNING] org.flywaydb:flyway-maven-plugin:4.0
[WARNING] *****************************************************************
```

axelfontaine:
The word parallel in this FAQ entry referred to clusters of machines and not Maven builds.

However I agree that having the ability to use Maven's parallel builds is a valuable feature.

axelfontaine:
This has been fixed as part of the Maven 3.0 conversion of the plugin.

Issue #1402, opened 2016-08-11
Make Maven 3 the new minimum supported version


axelfontaine:
Maven 2 has been EOL'd for 2.5 years now and hasn't seen a release since 2009.

See https://github.com/flyway/flyway/issues/468#issuecomment-239140982

PR #1652, opened 2017-05-29
Fixes gh #1305

davidkarlsen:
Looking at the guidelines at  https://cwiki.apache.org/confluence/display/MAVEN/Parallel+builds+in+Maven+3 and the code - there seems to be no shared state - and hence OK to set @threadSafe


axelfontaine:
@davidkarlsen Probably won't merge as we plan to implement #1402 in time for Flyway 5.0

davidkarlsen:
So what you are saying is that there will be no more 4.x releases?

axelfontaine:
This has now become obsolete as #1402 has been implemented and it includes this change.

PR #1661, opened 2017-06-05
Fixed #1627 Continue classpath scan on AccessControlException

### LINKED ISSUE #1627: Appengine & ClassPathScanner : throw AccessControlException since 4.1.0
###### What version of Flyway are you using?
4.2.0

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin, SBT plugin, ANT tasks)
- Flyway Java API 
- Maven dependencies
- Appengine (latest SDK, 1.9.51) Standard Environment (JAVA 7) . 

###### What database are you using (type & version)?
MySql 5.6.23

###### What operating system are you using?
Mac OSX El Capitan

###### What did you do?
Only append in the Development environment (as appengine libs are in a different classpath on Google Appengine server)

Basic configuration of flyway: 
- appengine Java helloworld project
- resources/db/migration/V1_init.sql
- fresh & clean database (only one table)
- Add java Flyway migration in a simple servlet:
```
  Flyway flyway = new Flyway();
  flyway.setDataSource(url,user,password);
  flyway.migrate();
```

- Result : Throw an exception : 

> org.flywaydb.core.api.FlywayException: Unable to scan for SQL migrations in location: classpath:db/migration

I've hunted the bug down to the location where it occurs : 
In fact, when the [ClassPathScanner](https://github.com/flyway/flyway/blob/master/flyway-core/src/main/java/org/flywaydb/core/internal/util/scanner/classpath/ClassPathScanner.java) try to `findResourceNames`, it hits the appengine sdk jar (in my case 'appengine-agentruntime.jar') that triggers a : 
java.security.AccessControlException: access denied ("java.io.FilePermission" "/Users/CtrlX/google-cloud-sdk/platform/google_appengine/google/appengine/tools/java/lib/impl/agent/appengine-agentruntime.jar" "read")

Basically, after some research, appengine disallow that type of action (opening appengine jar at runtime) and so it raise that FilePermission exception.

This is due to this [commit](https://github.com/flyway/flyway/commit/0baf57e73beb694e9c34e54ad377aa155c27a392) that change the logic of method `findResourceNames` by adding: 
```
    ...
    try {
       jarFile = new JarFile(url.toURI().getSchemeSpecificPart());
    } catch (URISyntaxException ex) {
       // Fallback for URLs that are not valid URIs (should hardly ever happen).
        jarFile = new JarFile(url.getPath().substring("file:".length()));
    }
    ...
```

It doesn't catch the AccessControlException, and the exception bubble up to [Scanner.scanForResources](https://github.com/flyway/flyway/blob/master/flyway-core/src/main/java/org/flywaydb/core/internal/util/scanner/Scanner.java), eventually transformed into a FlywayException.

For those how have the same issue : 
The bug appear in the release 4.1.0, so as a work around, you can use a previous version of flyway (like 4.0.3)

Jar that trigger FilePermission exceptions should be ignored or just warned.

FlywayException encapsulating an AccessControlException

johnmichelau:
Some environments (like the Google App Engine Jetty Dev Server) pull
in JAR's from the host system that are not allowed to be scanned at
runtime.  This causes an AccessControlException to be thrown by
java.security, which aborts the classpath scan completely and causes
Flyway to fail.  These JAR's will never have DB migrations in them,
so the failure is unnecessary.  Let's catch this exception and just
warn the user instead.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/util/scanner/classpath/ClassPathScanner.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/util/scanner/classpath/ClassPathScanner.java
@@ -32,6 +32,7 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.net.URLDecoder;
+import java.security.AccessControlException;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


flyway-core/src/main/java/org/flywaydb/core/internal/util/scanner/classpath/ClassPathScanner.java
```diff
@@ -192,38 +193,43 @@ private String toClassName(String resourceName) {
                 if ("file".equals(url.getProtocol())
                         && url.getPath().endsWith(".jar")
                         && !url.getPath().matches(".*" + Pattern.quote("/jre/lib/") + ".*")) {
-                    // All non-system jars on disk
-                    JarFile jarFile;
                     try {
```
johnmichelau: GitHub's diff did a pretty poor job of recognizing the actual change here.  I simply wrapped the existing code in a try-catch.

axelfontaine:
Thanks for the pull request. I ended up fixing this slightly differently.

Issue #1579, opened 2017-03-30
Easy Flyway test environment setup based on Docker


axelfontaine:
Currently it is tedious and time-consuming to set up all databases required to fully test Flyway. See https://flywaydb.org/documentation/contribute/code/database

It should be easier both for existing and new contributors to quickly fire up a complete test environment with all databases to be able to run the full set of Flyway tests.

Almost every single database vendor now maintains a set of official Docker images. This should be leveraged in a Docker Compose file. See https://docs.docker.com/compose/compose-file/

Additionally the existing test configuration should be adapted to reference the databases started in the individual containers.

Even if not all databases are included at first, this approach could debut with a subset and expand from there.

Who's up for the challenge and ready to step up and contribute?

kiview:
It might be beneficial to use https://github.com/testcontainers/testcontainers-java for this task. I'll look into the current integration test aproach for Flyway and see if I'll be able to contribute.

axelfontaine:
@kiview Any of these works for us. At the end of the day the most convenient one for our contributors (prio 1) with the fastest run times (prio 2) wins.

bsideup:
Hi all,

I'm a co-maintainer of TestContainers and here to help :)

I checked some tests and can verify that it should be easy to integrate TestContainers.

Let me know if you want me to contribute some basic TC support (like we did with Zipkin)

I'm also open to any questions you have :)

axelfontaine:
@bsideup Thanks! Yes, we're interested! Let's maybe start with a single database in the first PR to get warmed up and expand from there.

bsideup:
@axelfontaine sure :) Just point me at some test where you use a real database so I can play around and replace it with TestContainers' based instance.

axelfontaine:
@bsideup These ones for example: https://github.com/flyway/flyway/tree/master/flyway-core/src/test/java/org/flywaydb/core/internal/dbsupport/db2

axelfontaine:
@bsideup Did you have the chance to look into this yet?

bsideup:
Hi @axelfontaine,

Sorry, didn't have enough time because of job change, but I'll have it soon :)
FYI since db2 integration is mostly about preparing the DB, it will help if you provide a Dockerfile/image for it. Of course, I can do it for you, but it will speed it up :)

axelfontaine:
[...]

PR #1663, opened 2017-06-07
Support TestContainers JDBC connection string

inikolaev:
I would like to propose this change in order to support TestContainers JDBC connection string.

This way it will be possible to run database migration scripts against Dockerized database instance started by TestContainers.

Let me know if you have any questions.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/util/jdbc/DriverDataSource.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/util/jdbc/DriverDataSource.java
@@ -247,6 +247,10 @@ public class DriverDataSource implements DataSource {
      * @return The Jdbc driver.
      */
     private String detectDriverForUrl(String url) {
+        if (url.startsWith("jdbc:tc:")) {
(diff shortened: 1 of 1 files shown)


axelfontaine:
We are certainly interested in moving in this direction. I assume this PR is only the initial step and that a few more will follow. Could you detail your overall vision of what the final solution, when completed, would look like and which intermediate steps you see would be needed to get there?

axelfontaine:
Related to #1579

inikolaev:
With this change I merely wanted to add support for TestContainers. It's not about Flyway tests to use it internally at the moment.

I have a use case of being able to run database migration script against TestContainers database and using that database during development without using Docker Compose. This change along with https://github.com/testcontainers/testcontainers-java/pull/360 will allow to achieve this goal.

inikolaev:
@axelfontaine Is there anything I can do to facilitate with for this to be merged?

axelfontaine:
@inikolaev Can you sign the CLA so this can be merged?

inikolaev:
@axelfontaine Done!

Issue #1667, opened 2017-06-14
Flyway over SSH Tunnel does not work on macOS 10.12


apeschel:
###### What version of Flyway are you using?

Flyway 4.2.0 by Boxfuse

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin, SBT plugin, ANT tasks)

Command-line

###### What database are you using (type & version)?

MariaDB

mysql  Ver 15.1 Distrib 10.1.22-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2

###### What operating system are you using?

macOS Sierra 10.12.5

###### What did you do?
(Please include the content causing the issue, any relevant configuration settings, and the command you ran)

Set up an SSH Tunnel to a remote database:

```
ssh -L 3306:sqldb1:3306 jumpbox
```

Settings in a config file:
```
flyway.url=jdbc:mariadb://localhost:3306/XXXXX?serverTimezone=UTC
flyway.user=XXXXX
flyway.password=XXXXX
flyway.locations=com.xxx.db.migration,filesystem:workbench/src/main/resources/db/migration
flyway.jarDirs=workbench/target/scala-2.11/
```

Trying to test the connection to the DB:
```
flyway -X info -configFile=local_XXX_flyway.conf
```

Failure:

```
Flyway 4.2.0 by Boxfuse

DEBUG: Java 1.8.0_131 (Oracle Corporation)
DEBUG: Mac OS X 10.12.5 x86_64

DEBUG: Loading config file: /usr/local/Cellar/flyway/4.2.0/libexec/conf/flyway.conf
DEBUG: Unable to load config file: /Users/apeschel/flyway.conf
DEBUG: Unable to load config file: /Users/apeschel/projects/XXX-workbench/backend/flyway.conf
DEBUG: Loading config file: /Users/apeschel/projects/XXX-workbench/backend/local_XXX_flyway.conf
DEBUG: Using configuration:
DEBUG: flyway.configFile -> local_XXX_flyway.conf
DEBUG: flyway.jarDirs -> workbench/target/scala-2.11/
DEBUG: flyway.password -> ********************
DEBUG: flyway.url -> jdbc:mariadb://localhost:3306/XXX?serverTimezone=UTC&useSSL=false
DEBUG: flyway.user -> XXX
DEBUG: flyway.locations -> com.XXXanalytics.db.migration,filesystem:workbench/src/main/resources/db/migration
[...]

PR #1668, opened 2017-06-14
Fix SSH Tunnel Bug on macOS 10.12

apeschel:
This commit upgrades the mariadb-java-client library to fix the SSH
tunnel bug described in https://github.com/flyway/flyway/issues/1667

mariadb-java-client 1.4.5 has a bug that prevents a connection over an
SSH tunnel on macOS 10.12.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -135,7 +135,7 @@
         <version.h2>1.4.193</version.h2>
         <version.hsqldb>2.3.4</version.hsqldb>
         <version.jtds>1.3.1</version.jtds>
-        <version.mariadb>1.4.5</version.mariadb>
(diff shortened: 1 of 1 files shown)


apeschel:
Relevant issue: https://github.com/flyway/flyway/issues/1667

apeschel:
Didn't actually run the testing, since getting the old versions of Java requires an Oracle account and jumping through a bunch of hoops.

apeschel:
I took some time to create an Oracle account and get the JVMs set up. The tests passed successfully with this change.

apeschel:
Also confirmed that this build fixes the connectivity problem.

Issue #1671, opened 2017-06-15
Ignore common SQL*Plus commands


edwagner:
###### What version of Flyway are you using?
4.2.0

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin, SBT plugin, ANT tasks)
Command-line

###### What database are you using (type & version)?
Oracle

###### What did you do?
`flyway migrate` using scripts that are also meant to run through SQL*Plus.

###### What did you expect to see?
Flyway process the script without error.

###### What did you see instead?
Flyway passed some of the SQL*Plus commands on to the database which caused an exception.

I realize that it is documented that flyway doesn't honor SQL\*Plus commands, but it would be nice if Flyway didn't blow up when it encounters them. Flyway ignores `set define off;` but not other common SQL\*Plus commands.

umapathy08:
similar case for set serveroutput on, show errors, etc.

axelfontaine:
Duplicate of #1845

PR #1672, opened 2017-06-15
Ignore more SQL*Plus statements

edwagner:
This doesn't handle every SQL*Plus statement, but ignores some of the more common ones.

Closes flyway/flyway#1671

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/dbsupport/oracle/OracleSqlStatementBuilder.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/dbsupport/oracle/OracleSqlStatementBuilder.java
@@ -118,6 +118,13 @@ public class OracleSqlStatementBuilder extends SqlStatementBuilder {
 
     @Override
     public boolean canDiscard() {
-        return super.canDiscard() || statementStart.startsWith("SET DEFINE OFF");
+        return super.canDiscard() || isSqlPlusCommand();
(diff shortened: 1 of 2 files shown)


flyway-core/src/main/java/org/flywaydb/core/internal/dbsupport/oracle/OracleSqlStatementBuilder.java
```diff
@@ -118,6 +118,13 @@ protected String computeAlternateCloseQuote(String openQuote) {
 
     @Override
     public boolean canDiscard() {
-        return super.canDiscard() || statementStart.startsWith("SET DEFINE OFF");
+        return super.canDiscard() || isSqlPlusCommand();
+    }
+
+    private boolean isSqlPlusCommand() {
+        return statementStart.matches("SET\\s+(DEFINE|ECHO|TIMING|SERVEROUTPUT)\\s+(ON|OFF).*")
```
umapathy08: Please add "SHOW ERRORS" as well in this list

troosan: We can also ignore `prompt` (or even better, log the statement behind it.
And also ignore `spool`

axelfontaine:
Thanks for the pull request. We ended up implementing this differently. See #1845

PR #1674, opened 2017-06-15
Ignore exceptions when instructed to

edwagner:
Honor `WHENEVER SQLERROR CONTINUE` and `WHENEVER SQLERROR EXIT FAILURE`

Closes flyway/flyway#1673

PR #1677, opened 2017-06-19
Ignore `show error` SQL*Plus command

edwagner:
Add `SHOW ERROR` to the list of SQL*Plus commands being ignored.

Closes flyway/flyway#1671


axelfontaine:
How does this compare to #1672 ? Shouldn't one of the two be closed?

edwagner:
@axelfontaine Yes, it supersedes #1672. It's built on top of #1674 (which is a slightly different, but still somewhat related concern), so I wasn't entirely sure on how to manage all this. I'm not very familiar with this workflow, so please let me know how you recommend we proceed. Thanks for considering these PRs, and let me know if I can be doing anything in the future to make them easier to consume.

axelfontaine:
Thanks for the pull request. We ended up implementing this differently. See #1845

Issue #1675, opened 2017-06-16
ErrorOverrides: override errors into warnings and warnings into errors


umapathy08:
###### What version of Flyway are you using?

Flyway 4.1.1

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin, SBT plugin, ANT tasks)

Command-line

###### What database are you using (type & version)?

Oracle 11g

###### What operating system are you using?

RHEL 5.x

###### What did you do?
(Please include the content causing the issue, any relevant configuration settings, and the command you ran)

**1 - I tried to run the below stored procedure using flyway.** 

create or replace procedure sample1 is
BEGIN
  DBMS_OUTPUT.**1**put_line ('Hello Mars!');
END;
/
show errors

Flyway complained that "show errors" is not a valid sql command and failed. It completely makes sense. 

**2 - So I tried to remove show errors and ran the same script again second time. Sql file look like below:**

create or replace procedure sample1 is
BEGIN
  DBMS_OUTPUT.**1**put_line ('Hello Mars!');
END;
/

###### What did you expect to see?

I expect flyway to fail if there are any syntax errors in stored procedure.

I also understood that Oracle jdbc driver returns "warning" for the above execution. Is it a right ask to get flyway fail the migration as this is a syntax error stating the error code. As we can't fail the migration for all warnings, may be you can get the warning and check for the error code for syntax errors and throw an error explicitly. 

###### What did you see instead?

**For number 1 sql, And the output log looks like below:**

DEBUG: Locking table "test"."schema_version"...
DEBUG: Lock acquired for table "test"."schema_version"
DEBUG: Found statement at line 1: create or replace procedure sample1 is
BEGIN
  DBMS_OUTPUT.1put_line ('Hello Mars!');
END;

DEBUG: Found statement at line 6: show errors
Migrating schema "test" to version 1.15.0 - test1.sql
DEBUG: Executing SQL: create or replace procedure sample1 is
BEGIN
  DBMS_OUTPUT.1put_line ('Hello Mars!');
END;

WARNING: DB: Warning: execution completed with warning (SQL State: 99999 - Error Code: 17110)
DEBUG: Update Count: 0
DEBUG: Executing SQL: show errors
DEBUG: Update Count: 0
DEBUG: Rolling back transaction...
DEBUG: Transaction rolled back
ERROR: Migration of schema "test" to version 1.15.0 - test1.sql failed! Please restore backups and roll back database and code!
[...]

PR #1682, opened 2017-06-27
#1675: Forcing flyway to fail if there is a compilation error in oracle sql procedure

umapathy08:
This pull request is to address the issue #1675 . With this code, Flyway will throw an error and mark the execution as "failed" if there is a compilation failure in Oracle sql stored procedures. However, the stacktrace did not provide us the root cause of the compilation failure which needs to be enhanced further. 

Note: In Oracle SQL developer, the complete stacktrace is shown when we provide an option called "show errors". We need to find an alternative for this option in jdbc.


axelfontaine:
We decided to go for a more generic route instead with Error Handlers. See https://flywaydb.org/documentation/errorhandlers

Issue #1647, opened 2017-05-25
Flyway logs with too many info


umapathy08:
###### What version of Flyway are you using?

4.1.1

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin, SBT plugin, ANT tasks)

Flyway Command-line

###### What database are you using (type & version)?

Oracle 11g

###### What operating system are you using?

RHEL 5.x

###### What did you do?
(Please include the content causing the issue, any relevant configuration settings, and the command you ran)

Flyway logs are of three types as follows:

migrate :  without any logs in console
migrate -X:  with all sql statements line by line with execution status in console
migrate -q :  with only errors and warnings 

Issue: We need to see the execution results of the sql. So we ran flyway with "-X". But we have procedure and sql files with more than 1000 lines. So flyway prints all the sql files in the logs. 

###### What did you expect to see?

1. We don't want to see the sql file contents in the logs. It would be useful to see the content only if there is an error.
2. We would like to see only the execution results in the logs and not the sql statements.
eg: Update Count: 100

Can we get different switches for this functionality ? I have tried copying logback jars in the classpath but it did not have option to suppress the debug logs from flyway even if we set log level to "INFO" in logback.xml.

###### What did you see instead?

We see all the sql statements if we use -X.

axelfontaine:
We are happy with the current way things are. I don't feel the benefits of adding additional switches outweighs the complexity.

umapathy08:
I agree. But from DB developer perspective, it won't make sense to see 1000 lines of sql statements to find the execution results in a particular migration. I would suggest to have some basic level of logging(execution results) in info level. Your thoughts please ?

umapathy08:
@axelfontaine Would you be considering this request as a feature in 5.0 release ?

umapathy08:
Hi @axelfontaine Added a pull request to print the update count for flyway info in console logs so that developers can view the execution results of the particular migration. This will still be available in debug mode but the difference is that debug mode will also print the sql statements along with execution results.

PR #1683, opened 2017-06-27
#1647: added update count in Flyway info logs

umapathy08:
This pull request is to address #1647 . This code will print the update count for flyway info in console logs so that developers can view the execution results of the particular migration. This will still be available in debug mode but the difference is that debug mode will also print the sql statements along with execution results.


axelfontaine:
Thanks, but as already mentioned in #1647 we have no plans to change this.

PR #1687, opened 2017-06-30
Vertica specific tables

avmukilan:


avmukilan:
When will the next release be with all pull requests ?

PR #1695, opened 2017-07-06
Made logging level in doMigrateGroup() consistent with rest of file

sturzl:
The changed line paired with an info level statement so it seemed to be missing unless running at log level debug. All other logging in the class except this one line is also at the info level.


axelfontaine:
This log level is the intended one in order to only produce one line of output per migration in info. Works as designed.

PR #1717, opened 2017-07-24
Flyway Ant scans JAR files on the classpath

ttulka:
As a developer I want Flyway **Ant** to scan the **content of JAR files on the classpath**. With this functionality I don't have to unpack the artefacts and merge then into a location folder.

This works with a simple Ant build:
```
<?xml version="1.0" encoding="UTF-8"?>
<project name="FlywayTest" basedir="." default="main">

  <path id="lib.path">
    <fileset dir="lib" includes="flyway-*.jar,commons-logging*.jar,mysql-connector-java*.jar"/>
  </path>
  <taskdef resource="org/flywaydb/ant/antlib.xml" uri="antlib:org.flywaydb.ant" classpathref="lib.path"/>
  
  <target name="main" xmlns:flyway="antlib:org.flywaydb.ant">
    <path id="flyway.classpath">
      <pathelement path="flyway-migration" /><!-- *.sql, *.class and *.jar with migrations are located here -->
    </path>
    <property name="flyway.locations" value="classpath:" />
    
    <property name="flyway.schemas" value="test" />
    <property name="flyway.user" value="root" />
    <property name="flyway.password" value="1234" />
    <property name="flyway.url" value="jdbc:mysql://localhost:3306/?characterEncoding=utf-8" />
    
    <property name="flyway.validateOnMigrate" value="false" />
    
    <flyway:migrate />
  </target>
</project>
```
where flyway-migration contains:
```
**/*.sql
**/*.class
**/*.jar/**/*.sql
**/*.jar/**/*.class
```

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/util/scanner/classpath/ClassPathScanner.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/util/scanner/classpath/ClassPathScanner.java
@@ -182,7 +182,7 @@ public class ClassPathScanner implements ResourceAndClassScanner {
 
         boolean locationResolved = !locationUrls.isEmpty();
 
-        // Make an additional attempt at finding resources in jar files that don't contain directory entries
+        // Make an additional attempt at finding resources in jar files
         if (classLoader instanceof URLClassLoader) {
             URLClassLoader urlClassLoader = (URLClassLoader) classLoader;
             for (URL url : urlClassLoader.getURLs()) {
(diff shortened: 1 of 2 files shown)


flyway-core/src/main/java/org/flywaydb/core/internal/util/scanner/classpath/ClassPathScanner.java
```diff
@@ -199,16 +199,23 @@ private String toClassName(String resourceName) {
...
                     try {
-                        boolean directoryFound = false;
-                        Enumeration<JarEntry> entries = jarFile.entries();
-                        while (entries.hasMoreElements()) {
-                            if (entries.nextElement().isDirectory()) {
-                                directoryFound = true;
-                                break;
+                        // Look up resources from jar files within an Ant task
+                        boolean shouldLookupResources = classLoader.getParent() != null
+                                            && "AntClassLoader".equals(classLoader.getParent().getClass().getSimpleName());
```
axelfontaine: I am not sure this is the right place to detect the Ant ClassLoader. Why is this even necessary anyway? This fallback code is really just for broken jar files missing directory entries.

ttulka: Flyway Ant Tasks cannot lookup Java or SQL based migrations.  If my migrations come in different artifacts (as JARs) it would be nice when the Ant Task can take care of it and lookup the migrations from those JARs as well. So far I need to unpack the content of JARs into the classpath folder which brings more work on the build side and breaks the structure of the migrations (for example, my JARs pack different types of migrations in different namespaces).
This cannot be done by the AntClassLoader, the default class loader of Ant. So I used URLClassLoader and packed the AntClassLoader as its parent.

ttulka: I tried to explain it in the description of my commit with an example.

axelfontaine:
@ttulka We are planning to discontinue support for the Flyway Ant tasks, starting with Flyway 5.0. Would you be interested in taking over the development? I could give you access to a new `flyway/flyway-ant` repo for this.

ttulka:
Sure, I would like to!

axelfontaine:
@ttulka I have created a new repo (https://github.com/flyway/flyway-ant) and invited you as a team member. You can use this the flyway-ant module from this commit as your starting point: https://github.com/flyway/flyway/tree/ac266f42a5d365c1ffb4df0030b3281bb76508b8

For publishing please use a new group id such as org.flywaydb.ant and also please set up your own docs (you can cops the existing ones as a start) and let me know the new address so I can link to them. Let me know if you have questions or need my help for anything.

Cheers!

PR #1724, opened 2017-07-28
fix(gradle): make the flyway task classpath simpler

### LINKED ISSUE #1710: Resolve Gradle 4 deprecation warning
> Gradle now uses separate output directories for each JVM language, but this build assumes a single directory for all classes from a source set. This behaviour has been deprecated and is scheduled to be removed in Gradle 5.0.

I identified the cause is Flyway by using `gradle <task> -Dorg.gradle.deprecation.trace=true`
```
at org.gradle.api.internal.tasks.DefaultSourceSetOutput.getClassesDir(DefaultSourceSetOutput.java:79)
at org.gradle.api.internal.tasks.DefaultSourceSetOutput_Decorated.getClassesDir(Unknown Source)
at org.flywaydb.gradle.task.AbstractFlywayTask.runTask(AbstractFlywayTask.java:297)
```

That leads to the following [line](https://github.com/flyway/flyway/blob/0bfd37e2c4c7f1dfec9346c798464b5fb238fc86/flyway-gradle-plugin/src/main/java/org/flywaydb/gradle/task/AbstractFlywayTask.java#L306),
```java
URL classesUrl = sourceSet.getOutput().getClassesDir().toURI().toURL();
```

The fix is probably to iterate over each of the directories in `getClassesDirs()` and add them to `extraURLs`.

I haven't touched the plugin for 4 years, so best that I recall was I added this to support Java migrations. But code evolves so someone more familiar should verify that is the correct change.

jnizet:
The strategy to compose the classpath of the flyway task was to add the classes and resources directories of all sourceSets, and then add the compile, runtime, testCompile and testRuntime dependencies. So, unless I'm mistaken, this basically boils down to using the whole test runtime classpath, which is accessible right away from the gradle API.

In addition, the code causes a warning since gradle 4.0, which now uses one separate classes output directory per language in a source set, instead of a single classes directory per source set. Using the test runtime classpath also allows avoiding this warning.

fixes #1710

### DIFF
--- a/flyway-gradle-plugin/src/main/java/org/flywaydb/gradle/task/AbstractFlywayTask.java
+++ b/flyway-gradle-plugin/src/main/java/org/flywaydb/gradle/task/AbstractFlywayTask.java
@@ -17,18 +17,14 @@ package org.flywaydb.gradle.task;
 
 import org.flywaydb.core.Flyway;
 import org.flywaydb.core.api.FlywayException;
-import org.flywaydb.core.internal.util.ClassUtils;
 import org.flywaydb.core.internal.util.Location;
 import org.flywaydb.core.internal.util.StringUtils;
-import org.flywaydb.core.internal.util.UrlUtils;
 import org.flywaydb.gradle.FlywayExtension;
 import org.gradle.api.DefaultTask;
-import org.gradle.api.artifacts.ResolvedArtifact;
 import org.gradle.api.plugins.JavaPluginConvention;
-import org.gradle.api.tasks.SourceSet;
 import org.gradle.api.tasks.TaskAction;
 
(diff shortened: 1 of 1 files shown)


jnizet:
The test failure happens in core, that I didn't touch at all, so I don't think it has any relationship with my changes. I ran the flyway-gradle-plugin and flyway-gradle-plugin-largetest locally.

axelfontaine:
Thanks for the pull request. Could you sign the CLA to clear the path for merging?

jnizet:
CLA is signed now.

axelfontaine:
This has been fixed by a different PR in Flyway 5.0. Closing.

PR #1728, opened 2017-07-31
Use Windows CLASSPATH environment variable

jsross:
Added a check to see if CLASSPATH environment variable is set. If yes, then add the classpath variable to the java command classpath option


axelfontaine:
Thanks! Merged.

Issue #1729, opened 2017-08-01
Resource leak detected enabling Strict Mode on Android platform


mstarita:
###### What version of Flyway are you using?
3.2.1, 4.2.0

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin, SBT plugin, ANT tasks)
Java API in Android platform

###### What database are you using (type & version)?
Embedded SQLite

###### What operating system are you using?
Android v4.1.1

###### What did you do?
Execute a migration via API call with **Android Strict Mode enabled**.

###### What did you expect to see?
No resource leak

###### What did you see instead?
This stack trace from v3.2.1:
E/StrictMode( 1000): java.lang.Throwable: Explicit termination method 'close' not called
E/StrictMode( 1000): 	at dalvik.system.CloseGuard.open(CloseGuard.java:184)
E/StrictMode( 1000): 	at dalvik.system.DexFile.<init>(DexFile.java:80)
E/StrictMode( 1000): 	at org.flywaydb.core.internal.util.scanner.android.AndroidScanner.scanForClasses(AndroidScanner.java:73)
E/StrictMode( 1000): 	at org.flywaydb.core.internal.util.scanner.Scanner.scanForClasses(Scanner.java:72)
E/StrictMode( 1000): 	at org.flywaydb.core.internal.resolver.jdbc.JdbcMigrationResolver.resolveMigrations(JdbcMigrationResolver.java:73)
E/StrictMode( 1000): 	at org.flywaydb.core.internal.resolver.jdbc.JdbcMigrationResolver.resolveMigrations(JdbcMigrationResolver.java:43)
E/StrictMode( 1000): 	at org.flywaydb.core.internal.resolver.CompositeMigrationResolver.collectMigrations(CompositeMigrationResolver.java:126)
E/StrictMode( 1000): 	at org.flywaydb.core.internal.resolver.CompositeMigrationResolver.doFindAvailableMigrations(CompositeMigrationResolver.java:108)
E/StrictMode( 1000): 	at org.flywaydb.core.internal.resolver.CompositeMigrationResolver.resolveMigrations(CompositeMigrationResolver.java:94)
E/StrictMode( 1000): 	at org.flywaydb.core.internal.resolver.CompositeMigrationResolver.resolveMigrations(CompositeMigrationResolver.java:42)
[...]

PR #1730, opened 2017-08-01
Fix for issue #1729: Resource leak using Strict Mode on Android platform

mstarita:
This is a fix for the issue #1729


axelfontaine:
Thanks. Could you sign the CLA so this can be merged?

PR #1746, opened 2017-08-18
Fix # 1747 Could not find a valid Docker environment

dhoer:
This fixes the core test error "Could not find a valid Docker environment." Travis was also complaining (and still is complaining) about the amount of logging, so I set maven quiet attribute which cut verbosity a great deal.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,10 @@
-dist: precise
+dist: trusty
 sudo: required
 language: java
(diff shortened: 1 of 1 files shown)


axelfontaine:
This PR almost certainly doesn't fix the root cause (testcontainers failing to load the Docker environment due to confusion about Android libs on ClassPath)

Issue #1167, opened 2016-01-20
Clean on DB2 functions or procedures fails when two procedures have the same name


pswrdf:
Nobody get load on my comment there https://github.com/flyway/flyway/issues/464, so I create this issue

I faced with same error flyway 3.2.1: 
INFO: Database: jdbc:db2://localhost:50001/TEST1 (DB2/NT64 9.7)
Exception in thread "main" org.flywaydb.core.api.FlywayException: Unable to clean schema "DB2ADMIN"
Caused by: com.ibm.db2.jcc.am.SqlSyntaxErrorException: DB2 SQL Error: SQLCODE=-476, SQLSTATE=42725, SQLERRMC=DB2ADMIN.MYFUNCTION, DRIVER=3.63.75

REFERENCE TO FUNCTION function-name WAS NAMED WITHOUT A SIGNATURE, BUT THE FUNCTION IS NOT UNIQUE WITHIN ITS SCHEMA

axelfontaine:
Please include exacts steps on how to reproduce (small sample script that Flyway fails to clean).

Also please try the version from GitHub (flyway-core can easily be built from source with -DskipTests) and see if the problem still occurs.

axelfontaine:
Any news?

axelfontaine:
Closing due to lack of response from original reporter.

axelfontaine:
Fixed by #1722

PR #1750, opened 2017-08-21
pointing to specificname for procedures with same signature and support for cleaning temporary table too

franciskittu:
#1167 

P.S.: A little problem with build... I needed to comment the _scanForClassesSubPackage()_ of the **ClassPathScannerSmallTest** class to correctly create the build package. (the issue raised right after forking, so with no personal modification, I decided to do a pull request anyway)


axelfontaine:
@franciskittu `ClassPathScannerSmallTest` isssue should be resolved. Could you also add a testcase for this?

franciskittu:
@axelfontaine Testcase added as required. Please check it.  Thanks.

axelfontaine:
Merged. Thanks!

Issue #1733, opened 2017-08-02
Redshift JDBC 4.2 driver support


jeffreytli:
###### What version of Flyway are you using?
**4.2.0**

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin, SBT plugin, ANT tasks)
**Command-line**

###### What database are you using (type & version)?
**Redshift JDBC 4.2**

###### What operating system are you using?
**Mac OS 10.11.6**

###### What did you do?
(Please include the content causing the issue, any relevant configuration settings, and the command you ran)
**I tried running the following command: `flyway migrate` with the `username`, `password`, and `url` defined in flyway.conf.**

###### What did you expect to see?
I expected it to succeed and output:
```
Creating Metadata table: "PUBLIC"."schema_version"
Current version of schema "PUBLIC": << Empty Schema >>
Migrating schema "PUBLIC" to version 1 - Create person table
Successfully applied 1 migration to schema "PUBLIC" (execution time 00:00.062s).
```

###### What did you see instead?
**After running the above mentioned command, I get the following error: `ERROR: Unable to instantiate JDBC driver: com.amazon.redshift.jdbc42.Driver : com.amazon.redshift.jdbc42.Driver`.** 

###### Other notes
I downloaded the Redshift JDBC driver from [here](http://docs.aws.amazon.com/redshift/latest/mgmt/configure-jdbc-connection.html), and moved the 4.2 driver into the `drivers` directory. After digging around, I'm seeing that only 4.1 is currently supported. Are there any plans on supporting 4.2 in the near future? Or is this already supported and I'm just not configuring Flyway properly?

axelfontaine:
What is the output if you add `-X`?

axelfontaine:
This has now been implemented.

PR #1758, opened 2017-08-28
Fix for #1733. Support for Redshift JDBC 4.2

aleskiewicz:
Redshift has unified it's driver class name across different JDBC versions in version 1.2.1 (dated November 2016)

This was not always the case and there are different class names for drivers in versions prior to 1.2.1, used for different JDBC compatibility packages : http://docs.aws.amazon.com/redshift/latest/mgmt/configure-jdbc-connection.html.

Note about driver vaesions

This PR resolves this, by using the universal driver name as the main strategy (compatible with drivers 1.2.1+) and dedicated class names as fallback (by checking if that class exists on classpath) for older drivers.


axelfontaine:
Thank you very much for your pull request. We ended up implementing it slightly differently, defaulting to the `jdbc42` driver and then falling back to `jdbc41` and finally `jdbc40`.

Issue #1362, opened 2016-07-13
Header comment on stored procedure


andrey-poltavskiy:
###### What version of Flyway are you using?

4.0.3
###### What database are you using (type & version)?

Microsoft SQL Server 2014 - 12.0.4100.1 (X64) 
###### What operating system are you using?

Microsoft Windows [Version 10.0.10586]
###### What did you do?

-- =============================================
-- Description: ...
-- =============================================
CREATE PROCEDURE dwh.p_address_save () AS
BEGIN
...
END;
After compile this procedure in Managment Studio i see comment in DDL statment. After Flyway disapear.

maxiwheat:
+1 for this one, Can be worked around by making header comments inside procedures, but with a DB with hundreds of Stored Proc it's a little more annoying.

clabnet:
What version of Flyway are you using?
Flyway 4.1.0

What database are you using (type & version)?
Microsoft SQL Server 2012 (SP3) (KB3072779) - 11.0.6020.0 (X64) 
	Oct 20 2015 15:36:27 
	Copyright (c) Microsoft Corporation
	Standard Edition (64-bit) on Windows NT 6.2 <X64> (Build 9200: ) (Hypervisor)

What operating system are you using?
WIN 10 PRO EN

The same issue as above.

kzibreg:
What version of Flyway are you using?
Flyway 4.1.2
What database are you using (type & version)?
Microsoft SQL Server 2008 (SP4) - 10.0.6241.0 (X64)  Standard Edition (64-bit) on Windows NT 6.1 <X64> (Build 7601: Service Pack 1) (VM) 

What operating system are you using?
Windows 8.1 Enterprise

The same issue as above.

jaygurnani:
What version of Flyway are you using?
Flyway 4.0.3

What database are you using (type & version)?
Microsoft SQL Server 2014 (SP2) (KB3171021) - 12.0.5000.0 (X64) 

What operating system are you using?
Windows 10 

The same issue as above with stored procs

jaygurnani:
This seems to be deliberate by design? 

`    /**
     * @return Whether the current statement is only closed comments so far and can be discarded.
     */
    public boolean canDiscard() {
        return !insideAlternateQuoteStringLiteral && !insideQuoteStringLiteral && !insideMultiLineComment && !nonCommentStatementPartSeen;
    }`

[...]

PR #1767, opened 2017-09-11
Allow comment blocks before stored procs/views/etc in SQL Server

izep:
update SQLServerSqlStatementBuilder.java so that every comment is seen as a directive and sent to sql server. This will prevent comment headers from being removed. 
https://github.com/flyway/flyway/issues/1362


axelfontaine:
Well, except they're not directives. This muddies the semantics and doesn't properly fix the issue, nor does it contain a test case.

I am afraid this can't be merged.

izep:
Can you suggest a better route?

izep:
what about overriding canDiscard() instead?

PR #1776, opened 2017-09-20
Add support for Snowflake data warehouse

alexlevene:
Adding support for Snowflake (https://www.snowflake.net/). No decidedly unique requirements for this database product as compared to others, so the contribution steps were followed as-is.


mbreault:
Sorry for the late reply.  Awesome!  I will try it out this week.  Thank you so much.

jbadger-rp:
@alexlevene, do you have any pointers on how to build and include this as a dependency in another app? I'm feeling a bit green here, so I'd appreciate any help! so far I can get this to build by adding the following config to pom.xml:
```xml
<plugin>
    <artifactId>maven-assembly-plugin</artifactId>
    <version>3.1.0</version>
+   <configuration>
+       <tarLongFileMode>posix</tarLongFileMode>
+   </configuration>
</plugin>
```

and then running
```bash
mvn install -P-CommercialDBTest -P-CommandlinePlatformAssemblies -DskipTests=true
```

That said, this installs each flyway module separately instead of as one since it's setup as an aggregate maven project. My application isn't setup to consume multiple artifacts. I'd prefer to use this as a maven plugin, as defined in their [documentation](https://flywaydb.org/documentation/maven/). Is that possible since it'll be a local dependency?

alexlevene:
@jbadger-rp sorry, but I might be more green than you in this respect. After building the project, we ship the command line target for use by in our CI/CD processes, but the binary is just referenced in a shell script, not a direct dependency of other projects.

Btw, there is a SnowflakeDBTest profile you can use for install/build.

axelfontaine:
Superseded by #1914

PR #1781, opened 2017-09-27
remove redundant imports to fix #1545

### LINKED ISSUE #1545: Ensure Java 9 compatibility

jillesvangurp:
Looks like these imports were mistakenly added. Should fix java 9 compatibility.

Note. I was unable to build this project in any obvious way. Intellij seems fine with the change. Mvn simply fails. IMHO this should not break anything but probably a good idea for somebody to review this before merging. I don't have a huge amount of time to spend on fixing my environment. Sorry, just the way it is.

### DIFF
--- a/flyway-gradle-plugin/src/main/java/org/flywaydb/gradle/task/FlywayRepairTask.java
+++ b/flyway-gradle-plugin/src/main/java/org/flywaydb/gradle/task/FlywayRepairTask.java
@@ -16,7 +16,6 @@
 package org.flywaydb.gradle.task;
 
 import org.flywaydb.core.Flyway;
-import org.omg.CORBA.Object;
 
 /**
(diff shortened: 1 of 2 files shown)


axelfontaine:
Thanks for the hint! Weirdly enough IntelliJ did not even detect these as being unused, nor would it remove them automatically with organize imports.

Closing as this has been fixed in a different commit. Should have no effect on Java 9 compatibility though as imports do not affect runtime bytecode.

jillesvangurp:
@axelfontaine Somebody did report a class not found exception on this while running with java 9 (see #1545 ); might be worth backporting.

jezovuk:
I think it does affect Java 9 compatibility. Method run() returns an Object, so with (removed) CORBA Object import, that one is used instead of standard java.utli.Object. That is probably why IntelliJ did not report anything. Any chance of shipping this before 5.0.0?

Issue #887, opened 2014-11-24
Optional timeout on migrate process


PeeZu:
As improvement, migrate process should support a time-out property in order to prevent too long execution for each migration file. 
For instance:
On continuous delivery process, each changes are deployed on next environment if build success. the goal is to break the deployment process if the time-out occur => identify mistake due to the data volumetry not present in development environment (Thin database) comparing to the next deployment environment (Pre-Production, Staging etc...).

oriptipalti:
Would appreciate this as well.

For now, we have a workaround where we wrap the entire flyway run with a timeout script - we essentially kill the process if it's been running more than X seconds.
It's not the same behavior as in the suggestion and it's not very safe, but for us, it's better than nothing.

juliahayward:
Do you understand why your scripts are taking a long time? Eg. deadlocks, high database usage, ...? Would they be safe to interrupt mid-execution?

oriptipalti:
> Do you understand why your scripts are taking a long time? Eg. deadlocks, high database usage, ...? Would they be safe to interrupt mid-execution?

For us, it's been a lethal combination of heavy workload tasks like index creation and table locks we did not anticipate.

tbeauvais-imagine:
With millions of records, simple operations can take a very long time and even timeout. They are completed by the database eventually, yet Flyway times out not waiting for the result marking the migration as failed which requires manual intervention to fix.

Are you saying that the only way to increase the timeout for Flyway is to increase the timeout for all types of transitions?

manojsoft:
I'm also encountering this issue where my view creation is taking longer than 30 mins, it does eventually completes on database, but the flyway migration timeouts and tries to rollback the migration. 
I've been looking for a configurable parameter / default value for transaction timeout in flyway configs, but couldn't find it.

Here's the error log, from my flyway migration:
```
ERROR: Unable to rollback transaction
org.postgresql.util.PSQLException: This connection has been closed.
    at org.postgresql.jdbc.PgConnection.checkClosed(PgConnection.java:907)
    at org.postgresql.jdbc.PgConnection.rollback(PgConnection.java:914)
    at org.flywaydb.core.internal.jdbc.TransactionalExecutionTemplate.execute(TransactionalExecutionTemplate.java:71)
    at org.flywaydb.core.internal.database.postgresql.PostgreSQLAdvisoryLockTemplate.execute(PostgreSQLAdvisoryLockTemplate.java:56)
[...]

PR #1792, opened 2017-10-11
WIP: add support for migration timeout

mathieu-pousse:
Give a try to add support for migration timeout

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
@@ -325,12 +325,12 @@ public class Flyway implements FlywayConfiguration {
      */
     private String installedBy;
 
-
-
-
-
-
-
+    /**
+     * The migration timeout.
+     * <p>
+     *  {@code null} no timeout. (default: {@code null}).
+     */
+    private Long migrationTimeout;
 
 
     /**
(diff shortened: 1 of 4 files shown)


mathieu-pousse:
This is a first attempt to work on issue #887

The main drawback of this implementation is that we are forced to detach the migration on a separate Thread But as far as I can see, there is no other way in Java to timeout a callable.

Still need to polish and add some tests

Any remarks appreciated

mathieu-pousse:
Dany idea who thé ci faibles? Pas it working encore m'y changes?

mathieu-pousse:
@tjerkw personnaly, I don't especially need it anymore. 

We are also in a CD context and we cannot limit migrations by their execution delay.

What we ensure, is that the migration are run in a non blocking fashion, meaning the migration must not lock tables.
We also run the migrations before the application is actually deployed (this is mainly due to liveness probe from kubernetes)

With that, their is no reason to rollback any migration

tjerkw:
But some postresql migastions such als "alter table" lock the table by
default. You can't get around it. How do you solve that. What if you can't
get a lock?

MikielAgutu:
Is this still relevant? We've done some work to improve Postgres locking issues.

DoodleBobBuffPants:
Closing until further activity

PR #1796, opened 2017-10-16
Spring Autowire Supported Migrations

alexparlett:
PR which solves the issue raised in the comments of #1063 against latest version of flyway and spring.

Supports spring context based autowiring while maintaining the original pure jdbc template migrations.


axelfontaine:

Issue #1810, opened 2017-10-27
Support H2 MySQL compatibility mode


lukas-krecan-lt:
###### What version of Flyway are you using?

4.2.0

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin, SBT plugin, ANT tasks)

Java API

###### What database are you using (type & version)?

H2

###### What operating system are you using?

Mac OS

###### What did you do?
(Please include the content causing the issue, any relevant configuration settings, and the command you ran)

See this failing test https://github.com/flyway/flyway/pull/1811/files

###### What did you expect to see?

Not failing

###### What did you see instead?

```java
 org.flywaydb.core.internal.dbsupport.FlywaySqlException: 
Unable to drop "PUBLIC"."big_name"
----------------------------------
SQL State  : 42S02
Error Code : 42102
Message    : Table "big_name" not found; SQL statement:
DROP TABLE "PUBLIC"."big_name" CASCADE [42102-193]


	at org.flywaydb.core.internal.dbsupport.SchemaObject.drop(SchemaObject.java:82)
	at org.flywaydb.core.internal.dbsupport.h2.H2Schema.doClean(H2Schema.java:69)
	at org.flywaydb.core.internal.dbsupport.Schema.clean(Schema.java:148)
	at org.flywaydb.core.internal.command.DbClean$4.call(DbClean.java:184)
	at org.flywaydb.core.internal.command.DbClean$4.call(DbClean.java:181)
	at org.flywaydb.core.internal.util.jdbc.TransactionTemplate.execute(TransactionTemplate.java:75)
	at org.flywaydb.core.internal.command.DbClean.cleanSchema(DbClean.java:181)
	at org.flywaydb.core.internal.command.DbClean.clean(DbClean.java:130)
	at org.flywaydb.core.Flyway$3.execute(Flyway.java:1121)
	at org.flywaydb.core.Flyway$3.execute(Flyway.java:1117)
	at org.flywaydb.core.Flyway.execute(Flyway.java:1510)
	at org.flywaydb.core.Flyway.clean(Flyway.java:1117)
	at org.flywaydb.core.internal.dbsupport.h2.H2MigrationMediumTest.mysqlModeWithUppercase(H2MigrationMediumTest.java:63)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

PR #1811, opened 2017-10-28
H2 MySQL clean test

lukas-krecan-lt:


axelfontaine:
Closing. See comment on #1810 .

juliahayward:
Have tested with `com.h2database.h2` versions 1.4.193 through 199, and can confirm the test here passes at v1.4.199 but not older versions. Closing this as an H2 problem - Flyway 6.0.0 ships with the v1.4.199 H2 and should therefore be OK. If there are problems with v6.0, please open a new issue.

PR #1836, opened 2017-11-21
Add SQL vault feature: Allow encrypted SQL source file with AES128/AES256

GuillaumeSmaha:
Add sqlvault command to encrypt/decrypt sql source file easily
Add 3 parameters: askVaultPassword, vaultPasswordFile and vaultPassword
Add some StringUtils functions
Set private function in Main to protected
Add prompt for Vault Password

@axelfontaine I would like to add tests but you delete all tests in the commit: https://github.com/flyway/flyway/commit/3387a942a5eab7eb344df7d92b19524ff99c9b29

### DIFF
--- a/flyway-commandline/src/main/assembly/component.xml
+++ b/flyway-commandline/src/main/assembly/component.xml
@@ -23,6 +23,7 @@
             <outputDirectory></outputDirectory>
             <includes>
                 <include>flyway.cmd</include>
+                <include>sqlvault.cmd</include>
             </includes>
             <fileMode>644</fileMode>
             <filtered>true</filtered>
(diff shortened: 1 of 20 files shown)


axelfontaine:
Thanks for the pull request. What problem exactly does this solve from a user perspective? Please help me understand your scenario.

GuillaumeSmaha:
For example, I want to hide sensitive data contained in sql source file stored in a common git repository. So, with this feature I can encrypt the sql source file and anybody can't access to the content of the file.
During deployment, CD tool like Jenkins can provide the password to decrypt it and apply the migration.

axelfontaine:
Thanks for the explanation. I am unsure about adding this to the main Flyway distribution at this point as there are good external tools that can take care of that step as well.

axelfontaine:
@GuillaumeSmaha No worries, yes we plan to implement this in time for the 6.0.0 release and I wanted to keep it open for reference until then.

PR #1854, opened 2017-12-02
Adding comment surrounding usage of Oracle TNS Connection string

cdavid15:
Minor update to the example flyway.conf file bundled with the command line to explicitly show connections to oracle databases can be made using TNS names


axelfontaine:
Good idea. It makes sense to make this more explicit.

PR #1880, opened 2018-01-01
Add support for Google Cloud Spanner

olavloite:
This adds support for Google Cloud Spanner (https://cloud.google.com/spanner) to Flyway. The integration relies on the open source JDBC driver for Google Cloud Spanner (https://github.com/olavloite/spanner-jdbc) instead of the one provided by Google, as the official driver is read only.

# Context
The DDL syntax and possibilities of Cloud Spanner are close to standard SQL, albeit rather limited. Further specifics:
* Cloud Spanner does not support schemas. All databases contain a default schema named '' (emtpy string) and a system schema named INFORMATION_SCHEMA. The default schema may not be explicitly referenced, i.e. the statement `CREATE TABLE ``.TEST (ID INT64 NOT NULL, ...) PRIMARY KEY (ID);` will fail.
* Cloud Spanner does not support foreign keys, but tables can be interleaved in each other. When dropping interleaved tables, the child table must be dropped before the parent table.
* Tables can only be dropped after all secondary indices on the table have been dropped.
* DDL statements are very slow, especially if executed sequentially. The JDBC driver therefore supports both (automatic) batching and asynchronous execution of DDL statements.

# Example Application
I have created an example application showing the working of Flyway with Google Cloud Spanner. Google Cloud Spanner is only available in the Google Cloud, but you can opt for a free trial account a $300 credit for testing it.
The application can be found here: https://github.com/olavloite/spanner-flyway-example
An article describing the example can be found here: http://www.googlecloudspanner.com/2018/01/google-cloud-spanner-with-flyway.html


tuanng-cognite:
any plan to merge this?

luiccn:
Any plans on merging this? Very interested in using this the proper way :)

thinhha:
Ping @olavloite any plans to merge this please? :)

DoodleBobBuffPants:
Thanks for the PR! This is now available in the V7 beta which you can read more about [here](https://flywaydb.org/blog/flyway-7.0.0-beta).

PR #1887, opened 2018-01-06
Added support for Firebird database to flyway

andrevanzuydam:
This is still in testing phase as I have not completed dropping of stored procedures and resetting of generators but passes my basic tests which I need it for.  Also changed wording a bit for the schema stuff as Firebird is a non schema databases, the way I implemented may not be acceptable but would be easy to change, I had no joy returning a null for schema name, it evaluated to ""null"" so went with simple "null" as text return in Schema name over ride.  There will be other databases without schemas so instead of referring to schema now it says database?

Inside migrations:
Use --set term; to indicate stored procedure, I would not try migrate more than one multiline sql object at this time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Windows, macOS, Linux, Java and Android
 Maven and Gradle
 
 #### Supported databases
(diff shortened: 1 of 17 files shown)


andrevanzuydam:
Updating further functionality

PR #1888, opened 2018-01-07
Firebird Database Support

andrevanzuydam:
I have added support for the Firebird database engine:
Firebird is not a schema based database engine, I have attempted to indicate this with the messages but this may not be an acceptable change for the core flyway engine.  The actual implementation works fine with migrate and terminators using set term x; works correctly. I have also run tests with clean up which remove stored procedures,generators and tables correctly. I would want to update the documentation to this module as well as I know the Firebird database engine well.  I would be happy to implement the Interbase engine as another commit which would lean on this.

### DIFF
--- a/flyway-commandline/pom.xml
+++ b/flyway-commandline/pom.xml
@@ -80,13 +80,10 @@
             <groupId>org.xerial</groupId>
             <artifactId>sqlite-jdbc</artifactId>
         </dependency>
-
-
-
-
-
-
-
+        <dependency>
+            <groupId>org.firebirdsql.jdbc</groupId>
+            <artifactId>jaybird-jdk18</artifactId>
+        </dependency>
     </dependencies>
     <build>
         <resources>
@@ -139,12 +136,6 @@
                     </execution>
                 </executions>
             </plugin>
-
-
-
-
-
-
 
 
         </plugins>
@@ -235,15 +226,6 @@
                             </execution>
                         </executions>
                     </plugin>
-
-
-
-
-
-
-
-
-
 
 
                 </plugins>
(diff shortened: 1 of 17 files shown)


flyway-commandline/pom.xml
```diff
@@ -80,13 +80,10 @@
             <groupId>org.xerial</groupId>
             <artifactId>sqlite-jdbc</artifactId>
         </dependency>
-
```
axelfontaine: Please don't mess with whitespace lines like these. They are placeholders for keeping line numbers in sync between the community, pro and enterprise editions.

flyway-core/src/main/java/org/flywaydb/core/internal/database/Schema.java
```diff
@@ -148,7 +148,7 @@ public void clean() {
         try {
             doClean();
         } catch (SQLException e) {
-            throw new FlywaySqlException("Unable to clean schema " + this, e);
+            throw new FlywaySqlException("Unable to clean schema" + this, e);
```
axelfontaine: Please don't introduce formatting bugs.

axelfontaine:
Thanks for the pull request. Please see my initial comments.

andrevanzuydam:
Thanks, would you like me to revert and resubmit?

On Jan 7, 2018 6:49 PM, "Axel Fontaine" <notifications@github.com> wrote:

Thanks for the pull request. Please see my initial comments.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/flyway/flyway/pull/1888#issuecomment-355835733>, or mute
the thread

andrevanzuydam:
Going to fix this up

PR #1889, opened 2018-01-07
Firebird Database Support

andrevanzuydam:
@axelfontaine Removed my changes to the pom files and fixed the white space issue in Schema.java, that fell over when I was debugging, thanks for catching that.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/command/DbMigrate.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/command/DbMigrate.java
@@ -204,11 +204,19 @@ public class DbMigrate {
             currentSchemaVersion = infoService.current().getVersion();
         }
         if (firstRun) {
-            LOG.info("Current version of schema " + schema + ": " + currentSchemaVersion);
-
-            if (configuration.isOutOfOrder()) {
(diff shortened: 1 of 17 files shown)


flyway-core/src/main/java/org/flywaydb/core/internal/command/DbMigrate.java
```diff
@@ -204,11 +204,19 @@ private Integer migrateGroup(boolean firstRun) {
             currentSchemaVersion = infoService.current().getVersion();
         }
         if (firstRun) {
-            LOG.info("Current version of schema " + schema + ": " + currentSchemaVersion);
-
-            if (configuration.isOutOfOrder()) {
-                LOG.warn("outOfOrder mode is active. Migration of schema " + schema + " may not be reproducible.");
+            if (schema.getName() != "null") {
```
seregamorph: illegal equals check for strings

lvthillo: @andrevanzuydam ?

andrevanzuydam: I'm open to suggestions, I would need to build the project again to see what the implications were, Firebird does not have schemas

alextercete:
Superseded by #2376.

Issue #100, opened 2013-06-25
Clean tries to drop a function from an extension and fails


flyway:
_Original author: Martin.A...@gmail.com (May 11, 2012 10:11:31)_

We are using PostgreSQL 9.1 and are using the uuid-ossp extension. Clean tries to drop one of its functions. The extension has been installed manually as it required adding binaries to the DBMS.

Any way to work around this?

FlywayException: Error executing statement at line 64: DROP FUNCTION IF EXISTS &quot;public&quot;.&quot;uuid_nil&quot;() CASCADE
Occured in com.googlecode.flyway.core.migration.sql.SqlStatement in method execute, line number 78
Caused by org.postgresql.util.PSQLException: ERROR: cannot drop function uuid_nil() because extension uuid-ossp requires it
  Hint: You can drop extension uuid-ossp instead.
Occured in org.postgresql.core.v3.QueryExecutorImpl in method receiveErrorResponse, line number 2102

_Original issue: http://code.google.com/p/flyway/issues/detail?id=257_

imjoshholloway:
Just incase someone comes looking here and doesn't fancy jumping between the issues to figure out a work around:

Create a `beforeClean.sql` file in your migrations folder and add the required `DROP...` command into it.

For example, if like me, when doing `flyway clean -X` you get: `ERROR: cannot drop function uuid_nil() because extension uuid-ossp requires it` you would add the following into the `beforeClean.sql` file:

```sql
DROP EXTENSION IF EXISTS "uuid-ossp" CASCADE;
```

esetnik:
Wrong migration name format: beforeClean.sql(It should look like this: 1_2__Description.sql)

dustinevan:
@imjoshholloway, Beautiful summary. Thanks man.

tdonohue:
A complication to add into this:  Some Postgres extensions (e.g. uuid-ossp, pgcrypto) can only be dropped (DROP EXTENSION) by accounts with "superuser" privileges.  This means that @imjoshholloway's solution is great if your database owner also has "superuser" privileges, but won't work if you don't have "superuser" rights.

This unfortunately means that while normally a "clean" could be performed by a database owner, when extensions are installed, that database owner may not even have rights to remove them (unless they also have superuser privileges).

This seems to vary extension by extension though. The official docs say that some extensions should be manageable by database owners, while others require superuser privileges.
http://www.postgresql.org/docs/9.4/static/sql-createextension.html

cowwoc:
[...]

PR #1932, opened 2018-02-17
Do not delete types that extensions depend on when cleaning PostgreSql

yankee42:
I cannot use `clean` in one of my projects, because I am using the btree_gist extension. I alway get the following error:

> ERROR: cannot drop type gbtreekey4 because extension btree_gist requires it

This PR prevents that these types are dropped. Basically this is an extension to commit 444f077116b45260d7f99aa931d41ccb59553c8a which did the same for routines and views.


axelfontaine:
Related to #100

axelfontaine:
Thanks for the pull request! Merged. I have also taken the liberty to add you to the Hall of Fame page (flyway-5.1.0 branch).

PR #1933, opened 2018-02-26
Flyway 5.0 Crash on Android

benoitdevos:
When upgrading an existing Android app from Flyway 4.2 to 5.0.7, Flyway crashes with the following stack trace:
```
    Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.String.indexOf(java.lang.String)' on a null object reference
        at org.flywaydb.core.internal.database.DatabaseFactory.filterUrl(DatabaseFactory.java:204)
        at org.flywaydb.core.internal.database.DatabaseFactory.getJdbcUrl(DatabaseFactory.java:191)
        at org.flywaydb.core.internal.database.DatabaseFactory.createDatabase(DatabaseFactory.java:72)
        at org.flywaydb.core.Flyway.execute(Flyway.java:1634)
        at org.flywaydb.core.Flyway.info(Flyway.java:1321)
```
This is caused by `connection.getMetaData().getURL()` returning `null` when using SQLite and SQLDroidDriver as recommended. This is fixed in [this commit](../commit/545fd46c8dde422276e2f2215e3e1311a973c17d).

When fixed, another problem occurs:  Flyway complains that the SQLite version has to be at least 3.7.2, which:

- was not true with Flyway 4.
- is not compliant which [documentation on SQLite on Android](https://flywaydb.org/documentation/database/sqlite).

This is fixed in the [second commit](../commit/abc699d50f34cf6075d83c2d9a4cc559a19f6a7d).

Both these commits make Flyway working just fine on Android with SQLite and SQLDroidDriver.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/database/DatabaseFactory.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/database/DatabaseFactory.java
@@ -214,6 +214,9 @@ public class DatabaseFactory {
      * @return The filtered url.
      */
     static String filterUrl(String url) {
+        if (url == null) {
+            return "";
+        }
         int questionMark = url.indexOf("?");
         if (questionMark >= 0 && !url.contains("?databaseName=")) {
             url = url.substring(0, questionMark);
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/database/sqlite/SQLiteDatabase.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/database/sqlite/SQLiteDatabase.java
@@ -16,10 +16,10 @@
 package org.flywaydb.core.internal.database.sqlite;
 
 import org.flywaydb.core.api.configuration.FlywayConfiguration;
-import org.flywaydb.core.api.errorhandler.ErrorHandler;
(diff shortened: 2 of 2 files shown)


axelfontaine:
Thanks for the pull request. This is actually pretty frustrating as both of these issues are SQLDroid issues. And to add insult to injury they have already been fixed ages ago as well! All that is missing is a SQLDroid release. See https://github.com/SQLDroid/SQLDroid/issues/71

benoitdevos:
@axelfontaine Sorry I just wanted to provide a patch that let Flyway and Android enthusiasts to seamlessly  migrate to version 5, but wording of this PR was not adequate. I do agree that this is a work around SQLDroid "weakness".
I will consider pushing an official fork of SQLDroid...

axelfontaine:
No need to be sorry. I am frustrated with SQLDroid, not with your pull request :-)

axelfontaine:
Closing as the null check has been implemented as part of a different commit. See https://github.com/flyway/flyway/blob/master/flyway-core/src/main/java/org/flywaydb/core/internal/database/DatabaseFactory.java#L226

Issue #2035, opened 2018-06-05
PostgreSQL: Restore role to its original value between migrations instead of resetting it


danieljamesscott:
###### Which version and edition of Flyway are you using?

5.1.1

###### If this is not the latest version, can you reproduce the issue with the latest one as well?
(Many bugs are fixed in newer releases and upgrading will often resolve the issue)

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin)

Java API

###### Which database are you using (type & version)?

PostgreSQL 9.6

###### Which operating system are you using?

Linux

###### What did you do?
(Please include the content causing the issue, any relevant configuration settings, the SQL statement that failed (if relevant) and the command you ran.)

I need to be able to run 'set role $rolename' for all flyway migrations, including the initial metadata table creation. This is to ensure that the created objects are owned by the 'correct' user. We are using vault for database credential management, so the user running the migrations is dynamic and will change over time, leading to inconsistent ownership of objects.

###### What did you expect to see?

Possibly a flyway configuration option, or callback support to call 'set role'. I'm not sure if there's a callback which is run before the initial metadata table creation?

###### What did you see instead?

I saw a few issues related to this, including a call to 'RESET ROLE' which appears to be conflicting with my attempts to use connectInitSql on the pool, or the beforeMigration callback.

https://github.com/pradheeps/flyway/commit/5671442ea36fcf595ef8221271ba0622156e0b8f#diff-082a3414ab12680389b61424a632705cR90

axelfontaine:
This should be achievable using a combination of callbacks.

For the initial schema history table creation, use the `beforeMigrate.sql` callback and then for the individual migrations use the `beforeEachMigrate.sql` callback.

danieljamesscott:
Thanks, but would it be possible to add this in a cleaner way? I believe that this will become more of a problem as people start using dynamic users more.

axelfontaine:
I am not sure a dedicated setting would be pulling its conceptual weight at this point, especially considering the easy workaround available.

But the future will tell. If this does prove to be a major pain point down the road, we'll gladly reconsider this decision.

danieljamesscott:
Sure. Thanks.

Can we make it possible to disable the 'RESET ROLE' call? It seems a little strange to do a hard reset. If someone has changed role, then it should be their responsibility to change it back when they're done, right?

axelfontaine:
[...]

PR #2036, opened 2018-06-06
Add configuration option to restore original state

danieljamesscott:
This is a new feature to add a configuration option to toggle off the restoreOriginalState call which is executed at various places throughout the code.

The existing code causes a problem when trying to solve an issue which is encountered when we perform migrations using different users. The database objects end up being owned by the user which ran the migration which we attempt to solve by setting the role used by the migration. The RESET ROLE call breaks the attempted fix.

I have attempted to work around this by using beforeMigrate and beforeEachMigrate callbacks, but this does not work for the schema history table (This broke around Version 5.0.7).

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
@@ -792,6 +792,9 @@ public class Flyway implements Configuration {
         return configuration.isSkipDefaultCallbacks();
     }
 
+    @Override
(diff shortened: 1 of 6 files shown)


axelfontaine:
Thanks for the PR. However as I stated in #2035 I don't feel adding this property makes sense at this point. Also should we decide to add it one day, the change would be much bigger as all clients and the docs on the website would need to be updated as well.

The workaround I suggested should work with 5.1.1. Please debug and comment on the original issue if for some reason it doesn't.

PR #2042, opened 2018-06-14
Workaround a bug in windows cmd

dar-brett:
When calling flyway using System.Diagnostics.Process in .Net to create a database as part of my test suite setup I encountered a bug in Windows CMD with how %~dp0 is resolved, a bit of information and the solution I borrowed from can be found here: https://stackoverflow.com/a/26851883


axelfontaine:
Hmmm interesting... I had no idea! Thanks for the fix!

Issue #2128, opened 2018-08-31
Add an 'ignorePendingMigrations' configuration option


yjo:
Add an 'ignorePendingMigrations' option to suppress validation errors caused by the availability of not-yet-applied migrations.

My use case is for a CI/CD pipeline - while developers are working on a feature branch they should be able to create and modify migrations for that feature.  However, we should be able to detect and break the build when changes are made to versioned migrations that have already been applied to (e.g.) our production database.  Running `flyway validate` with `ignorePendingMigrations=true` would support this scenario and be consistent with the existing `ignore*Migrations` options.

axelfontaine:
This is already possible by passing `-target=current` to the `flyway validate` command.

yjo:
@axelfontaine `-target=current` doesn't seem to work in exactly the same way in the presence of out-of-order migrations.  E.g. if I run `flyway -target=current -ignoreIgnoredMigrations=true -ignoreFutureMigrations=true -outOfOrder=true validate` then I can still get "Detected resolved migration not applied" errors.  This happens when running with a migration set with a new migration ordered before the current DB version, which can occur when team members work on feature branches that overlap in time.  When running with the proposed `ignorePendingMigrations=true` option the validation works as expected.

yjo:
Hmm, the right configuration of my use case seems to be to use `-ignoreIgnoredMigrations=true -outOfOrder=false` which isn't super intuitive.  I think one source of difficulty is that the documentation mostly assumes a very linear, serialised version history (corresponding to application release versions), and its examples don't often apply to scenarios where a team creates migrations in a distributed, parallel way (as application features are developed).

yjo:
Versioned migrations aside, I can't find any combination of existing flags that will pass validation with changes to a repeatable migration.  It seems that might only be achievable with an `ignorePendingMigrations` option, @axelfontaine?

axelfontaine:
@yjo Thanks for the clarifications. I am starting to warm to the idea of adding this. Reopening.

PR #2129, opened 2018-08-31
Add an 'ignorePendingMigrations' configuration option

yjo:
This adds an 'ignorePendingMigrations' option to suppress validation errors caused by the availability of not-yet-applied migrations (feature request #2128).

My use case is for a CI/CD pipeline - while developers are working on a feature branch they should be able to create and modify migrations for that feature.  However, we should be able to detect and break the build when changes are made to versioned migrations that have already been applied to (e.g.) our production database.  Running `flyway validate` with `ignorePendingMigrations=true` supports this scenario and is consistent with the existing `ignore*Migrations` options.

### DIFF
--- a/flyway-commandline/src/main/assembly/flyway.conf
+++ b/flyway-commandline/src/main/assembly/flyway.conf
@@ -210,6 +210,13 @@ flyway.url=
 # true to continue normally, false to fail fast with an exception. (default: false)
 # flyway.ignoreIgnoredMigrations=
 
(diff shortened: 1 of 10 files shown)


axelfontaine:
Not necessary. See #2128

axelfontaine:
Merged. I had to make some minor adjustments as you had missed a few places where this property had to be carried over (it's not easy to catch them all). Thanks for the PR and thanks for detailing your use case in #2128 .

Issue #1807, opened 2017-10-25
Disable recursive filesystem scanning down hidden directories


nelvadas:
What version of Flyway are you using?

4.2.0
Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin, SBT plugin, ANT tasks)

Command line
What database are you using (type & version)?

Postgres 9.6
What operating system are you using?

Containers:
Migration container Alpine Linux
DB container: centos
What did you do?

(Please include the content causing the issue, any relevant configuration settings, and the command you ran)
Trying to migrate apply a simple SQL Migration on a postgres DB from openshift, The data SQL are provided by a mounted volume .Files seems replicated at runtime and the migration failed with the error

In my /var/flyway/data I provided only one file.

When connecting to the migraiton container and run the command manually in debug mode i have htis

$/opt/flyway/4.2.0 # ./flyway info -user=root -password=Europa01# -url=jdbc:postgresql://beosbank-posgres-db-europa/beosbank-europa -locations=filesystem:/var/
flyway/data -X
Flyway 4.2.0 by Boxfuse

DEBUG: Java 1.8.0_131 (Oracle Corporation)
DEBUG: Linux 3.10.0-514.6.1.el7.x86_64 amd64

DEBUG: Loading config file: /opt/flyway/4.2.0/conf/flyway.conf
DEBUG: Unable to load config file: /root/flyway.conf
DEBUG: Unable to load config file: /opt/flyway/4.2.0/flyway.conf
DEBUG: Using configuration:
DEBUG: flyway.jarDirs -> /opt/flyway/4.2.0/jars
DEBUG: flyway.password -> *********
DEBUG: flyway.url -> jdbc:postgresql://beosbank-posgres-db-europa/beosbank-europa
DEBUG: flyway.user -> root
DEBUG: flyway.locations -> filesystem:/var/flyway/data
DEBUG: Adding location to classpath: /opt/flyway/4.2.0/drivers/derby-10.12.1.1.jar
DEBUG: Adding location to classpath: /opt/flyway/4.2.0/drivers/derbyclient-10.12.1.1.jar
DEBUG: Adding location to classpath: /opt/flyway/4.2.0/drivers/greenplumdriver-5.1.1.jar
DEBUG: Adding location to classpath: /opt/flyway/4.2.0/drivers/h2-1.4.193.jar
DEBUG: Adding location to classpath: /opt/flyway/4.2.0/drivers/hsqldb-2.3.4.jar
[...]

PR #2142, opened 2018-09-15
Add option to ignore hidden directories on recurse

olib963:
I think I have done this correctly, this should address https://github.com/flyway/flyway/issues/1807. I could not see any test files in the source (please let me know if I am missing something) so had to just test it out myself.


olib963:
Would you like this change also applied to classpath scanning? Or is file system enough?

axelfontaine:
Thanks for the pull request. We decided to enable this behavior by default without a switch.

Issue #2167, opened 2018-10-06
Allow versioned migration names like V12.sql (no separator, no description)


axelfontaine:
This should be possible for both regular and undo versioned migrations

PR #2172, opened 2018-10-13
#2167 Versioned Migrations need not have separator as well as description

satishthulva:


axelfontaine:
Thanks for the pull request. We decided to implement this differently.

PR #2199, opened 2018-11-06
Fixes #2198: fixed resource scanning for felix 6.0.1

### LINKED ISSUE #2198: Flyway fails to find resources in Felix 6.0.1 environment
5.2.1

Java API

###### Which database are you using (type & version)?
Postgres

Linux

###### What did you do?
Upgraded Felix to 6.0.1

Flyway worked as before

Flyway fails to find resources

The reason is felix changes resource URL schemas, they look now like this:

```
"bundle://e3a74e5a-af1f-46f0-bb53-bc5fee1b4a57_145.0:1/db/connect-migrations"
```

The bundle id is 145 in the example above.

renat-sabitov:
@axelfontaine 
you will need to upgrade maven-bundle-plugin to 3.2.0 to handle java 8
```
                 <plugin>
                     <groupId>org.apache.felix</groupId>
                     <artifactId>maven-bundle-plugin</artifactId>
-                    <version>2.3.7</version>
+                    <version>3.2.0</version>
                 </plugin>
```

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/scanner/classpath/OsgiClassPathLocationScanner.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/scanner/classpath/OsgiClassPathLocationScanner.java
@@ -19,11 +19,13 @@ import org.osgi.framework.Bundle;
 import org.osgi.framework.FrameworkUtil;
 
 import java.net.URL;
+import java.util.AbstractMap;
+import java.util.Arrays;
 import java.util.Enumeration;
+import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.TreeSet;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 /**
  * OSGi specific scanner that performs the migration search in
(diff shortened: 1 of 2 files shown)


flyway-core/src/main/java/org/flywaydb/core/internal/scanner/classpath/OsgiClassPathLocationScanner.java
```diff
@@ -58,21 +56,17 @@
...
-    private long getBundleId(String host) {
-        final Matcher matcher = bundleIdPattern.matcher(host);
-        if (matcher.find()) {
-            return Double.valueOf(matcher.group()).longValue();
-        }
-        throw new IllegalArgumentException("There's no bundleId in passed URL");
+        final String path = locationUrl.getPath();
+        final String host = locationUrl.getHost();
+        return Arrays.stream(currentBundle.getBundleContext().getBundles())
+                .map(b -> new AbstractMap.SimpleEntry<>(b, b.getEntry(path)))
```
renat-sabitov: Note that this implementation relies on jars have directory entries. I believe all modern tools produce jars with directory entries, and other tools like Spring rely on them being present, so it shouldn't be a huge issue.

axelfontaine:
Thanks for the pull request. We ended up fixing it differently differently. See comment on #2198.

PR #2203, opened 2018-11-07
Merge Flyway 5.2.2 fixes

axelfontaine:


flyway-core/src/main/java/org/flywaydb/core/internal/database/base/SchemaObject.java
```diff
@@ -32,12 +32,12 @@
...
      * The database-specific support.
      */
-    protected final Database database;
+    protected final D database;
 
     /**
      * The schema this table lives in.
      */
-    protected final Schema schema;
+    protected final S schema;
```
vrnsky: Hm...one letter type is not good in my opinion. Maybe should use ```SchemaExtended``` or ```SchemaChild```

flyway-core/src/main/java/org/flywaydb/core/internal/database/derby/DerbyDatabase.java
```diff
@@ -64,19 +63,11 @@ protected DerbyConnection getConnection(Connection connection
 
     @Override
     public final void ensureSupported() {
-        String version = majorVersion + "." + minorVersion;
+        ensureDatabaseIsRecentEnough("Derby", "10.11.1.1");
```
vrnsky: maybe make constant value ```Derby``` and ```10.11.1.1```

Issue #2236, opened 2018-12-07
canExecuteInTransaction is not respected by JavaMigration


logicbomb:
###### Which version and edition of Flyway are you using?
5.2.4

###### If this is not the latest version, can you reproduce the issue with the latest one as well?
(Many bugs are fixed in newer releases and upgrading will often resolve the issue)

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin)
Java API

###### Which database are you using (type & version)?
PostgresSQL

###### Which operating system are you using?
Ubuntu 18.10

###### What did you do?
(Please include the content causing the issue, any relevant configuration settings, the SQL statement that failed (if relevant) and the command you ran.)
Write a java migration that must execute outside of a transaction - `alter type public.my_type add value 'new_value';`

###### What did you expect to see?
No error

###### What did you see instead?
A postgres error

The issue comes from the fact that the JavaMigrationExecutor doesn't forward the migration's `canExecuteInTransaction` value, which can be seen here:
https://github.com/flyway/flyway/blob/master/flyway-core/src/main/java/org/flywaydb/core/internal/resolver/java/JavaMigrationExecutor.java#L68

The method should be
```java
    @Override
    public boolean canExecuteInTransaction() {
        return this.javaMigration.canExecuteInTransaction();
    }
```

Also, I had a tough time debugging it b/c the reflection code that instantiates the migration hides inner exceptions, and is difficult to find.  I'm working on a PR that fixes the issue, and makes reflection errors easier to detect.

logicbomb:
Here's the code that makes debugging difficult: https://github.com/flyway/flyway/blob/master/flyway-core/src/main/java/org/flywaydb/core/internal/util/ClassUtils.java#L128-L164

axelfontaine:
Thanks for catching this and investigating. Truly appreciated! Fixed.

pjroth:
Is it possible to get a 5.X.Y release including this fix since 6.X.Y isn't yet ready for production?

PR #2237, opened 2018-12-07
Issue 2236 fix can execute in transaction

logicbomb:
A fix for https://github.com/flyway/flyway/issues/2236


axelfontaine:
Thanks for the pull request. We decided to fix this differently.

Issue #2293, opened 2019-02-01
Atomic DDL support for MySQL 8


lejtmany:
First of all, thank you for flyway. It's a great tool!

###### Which version and edition of Flyway are you using?
5.2.1

###### If this is not the latest version, can you reproduce the issue with the latest one as well?
(Many bugs are fixed in newer releases and upgrading will often resolve the issue)

yes

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin)

command line

###### Which database are you using (type & version)?
mysql 8.13

###### Which operating system are you using?
ubuntu 18.04

###### What did you do?
(Please include the content causing the issue, any relevant configuration settings, the SQL statement that failed (if relevant) and the command you ran.)
ran a migration with multiple DDL statements. One of which failed.

###### What did you expect to see?

I expected the previous DDL statements that had already executed to be rolled back as mysql 8 now supports atomic DDL transactions https://dev.mysql.com/doc/refman/8.0/en/atomic-ddl.html . Flyway however appears to no differentiate between mysql versions in MySQLDatabase.supportsDdlTransactions().

###### What did you see instead?
The transaction was not rolled back and instead the transaction was marked as failed.

lejtmany:
Made a PR:
https://github.com/flyway/flyway/pull/2294

lejtmany:
Ah, sorry my misunderstanding of the docs. I see now it says explicitly

"DDL statements, atomic or otherwise, implicitly end any transaction that is active in the current session, as if you had done a COMMIT before executing the statement. This means that DDL statements cannot be performed within another transaction, within transaction control statements such as START TRANSACTION ... COMMIT, or combined with other statements within the same transaction."

Apologies for wasting your time.

PR #2294, opened 2019-02-01
supports atomic ddl for mysql > 8

lejtmany:
mysql 8 now supports atomic ddl. 
https://dev.mysql.com/doc/refman/8.0/en/data-type-defaults.html


lejtmany:
response to https://github.com/flyway/flyway/issues/2293

Issue #2178, opened 2018-10-18
PostgreSQL 11 support


axelfontaine:

whaleyr:
I just tried 11 today. I was able to run the `migrate` tasks just fine. However, I did a `clean` and it borked when trying to query `pg_proc` for column `pgisagg` which has been removed between 10 and 11.

Keep up the good work, I'll be glad to see 6.0.0 when it's ready.

markusheiden:
Since the 6.0.0 seems to be far away: Are there any plans for a backport?

ragged:
We also have a big need for the 6.0.0 release, as a lot of upgrades are currently postponing, due to the lack of flyway support. How likely is 6.0.0 is on time?

ssonnenberg:
Yes, we are waiting. PostgreSQL 11 is available since November, so if anything is blocking
this small fix than please give us some details.

axelfontaine:
We have now pushed the out as part of **Flyway 6.0.0-beta**. See https://flywaydb.org/blog/flyway-6.0.0-beta

As part of this release, we are also making some changes to our support timeline for new database versions. Previously we added support for a new database version once it reached GA. We have now changed this and started adding support for a new database version once it reaches beta. The first database versions to benefit from this are SQL Server 2019 CTP and MariaDB 10.4 beta. Support for both of these is already included as part of 6.0.0-beta.

Going forward we will add support for PostgreSQL 12 as soon as a first beta is released, which will be many months before GA, so that a wait like the one for PostgreSQL 11 won't happen again.

ssonnenberg:
Great news. Will you provide maven coordinates to enable wider testing?

axelfontaine:
@ssonnenberg https://flywaydb.org/documentation/maven/

dahu33:
PostgreSQL 11 was released 10months ago, PostgreSQL 12 is in beta 2 and there is still no stable version of flyway supporting them. I don't understand the strategy of blocking people and companies by not adding compatibility in Flyway v5, especially since the required code change is trivial...

douglasrlee:
So, is the plan here just to wait for pg 12 and flyway 6?

SebHeuze:
Hi, is 14/08/19 still the release date planned for 6.0.0 release ?

PR #2299, opened 2019-02-08
Update PostgreSQLDatabase.java

ssonnenberg:
Could not see any problems during our internal tests.
Schema migrations work as expected with same results.


axelfontaine:
Duplicate of #2178

PR #2322, opened 2019-03-08
log with debug level in case sqlState is 00000

dwi-di:
SQLState 00000 indicates a **successful completion** of a query.
In this case the warnings are not really warnings but mere debug information on what was executed.

e.g.
DB: Run command: ALTER TABLE my_table DROP CONSTRAINT IF EXISTS my_constraint;

on log-level INFO it should be sufficient to see which migration is currently running. logging out all the individual queries that have been executed (successfully) should be only done for debugging purposes.


axelfontaine:
This works as expected. You can use error overrides if you wish to customize the default behavior of the database. See https://flywaydb.org/documentation/erroroverrides

PR #2325, opened 2019-03-12
Update H2Schema to work with DATABASE_TO_UPPER=false

josephlbarnett:
With H2 1.4.198 and DATABASE_TO_UPPER=false, flyway fails being
unable to find the (lowercase) schemata table and related columns.

Updating to use uppercase versions of the same makes things work.
With H2 1.4.197, this change was not necessary for some reason though?


josephlbarnett:
Need anything further from me to move this forward?

amyDoxy:
still have this issue with H2 v2.1.210 & Flyway v7.0.0

PR #2334, opened 2019-03-21
Enable configFiles configuration for maven plugin

Saljack:
I am not able use multiple config files from pom.xml. After this commit is posible set multiple config files like:
```
<configFiles>
  <configFile>flyway-dev.conf</configFile>
  <configFile>flyway.conf</configFile>
</configFiles>
```


axelfontaine:
Good catch! Thank you for investigating and fixing!

Issue #2048, opened 2018-06-21
Oracle view creation fails while execute the SQL script via Flyway


Sajith1988V:
###### Which version and edition of Flyway are you using?

Using Flyway Flyway Community Edition 5.1.1 by Boxfuse

###### If this is not the latest version, can you reproduce the issue with the latest one as well?
(Many bugs are fixed in newer releases and upgrading will often resolve the issue)

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin)
Command-line

###### Which database are you using (type & version)?
Oracle 12.2.1, Ojdbc 8

###### Which operating system are you using?
Windows & Unix

###### What did you do?
(Please include the content causing the issue, any relevant configuration settings, the SQL statement that failed (if relevant) and the command you ran.)

CREATE OR REPLACE VIEW PA_SUMMARY_VIEW AS
    WITH
        FUNCTION Test_funct (
            CL_ID   IN VARCHAR2,
            P_DIN   IN VARCHAR2
        ) RETURN NUMBER AS
            V_O_AMT   NUMBER(8,2);
        BEGIN
.
.
.
.
This is a new feature of Oracle 12 C to allow functions/procedures inside a with clause. I am trying to create a view out of this with clause query. This executes successfully when I run the script from SQL* Plus or SQL Developer. But this fails when executed via Flyway 5.1.1 Open source

###### What did you expect to see?

View created successfully. 

###### What did you see instead?

Encountered the below error

**SQL State  : 65000
Error Code : 6553
Message    : ORA-06553: PLS-103: Encountered the symbol "end-of-file" when expecting one of the following:

   := ; not null default character**

axelfontaine:
Please share the full statement Flyway fails to parse correctly, so we can make sure you case is covered.

Sajith1988V:
Here is the full view query:



CREATE OR REPLACE VIEW PA_SUMMARY_VIEW AS
    WITH
-- Total Amount Function    
        FUNCTION WITH_TOT_AMT (
            CL_ID   IN VARCHAR2,
            P_DIN   IN VARCHAR2
        ) RETURN NUMBER AS
            V_O_AMT   NUMBER(8,2);
        BEGIN
            SELECT
                SUM(A.TOTAL_AMT_PAID) TOTAL_AMT_$
            INTO V_O_AMT
            FROM
                table1 A,
                table2 B
            WHERE
                A.CARRIER_ID = 15
                AND A.CEXP_REC_ID = 4
[...]

PR #2456, opened 2019-08-01
Very Quick Fix for #2455

### LINKED ISSUE #2455: Regexp in #2048 is to broad
5.2.4 Enterprise

Maven

###### Which database are you using (type & version)?
Oracle 12.2

Linux (windows on developer machines)

###### What did you do?
Create View statement fails if using "AS WITH" but not followed by an PL/SQL statement:

```sql
CREATE OR REPLACE VIEW V_TEST_1
AS
WITH data_list AS (
    SELECT 1 AS ID FROM DUAL
)
SELECT ID
  FROM data_list;

CREATE OR REPLACE VIEW V_TEST_2
AS
SELECT 1 AS ID FROM DUAL;

-- another statement
```

Successful migration

ORA-00933: SQL command not properly ended

## Analysis

The fix for #2048 is incomplete, the Regexp is too broad:

It should only apply to WITH followed by PROCEDURE or FUNCTION, not a regular SQL statement.

Instead the syntax should be:

```java
"^CREATE(\\s+OR\\s+REPLACE)?(\\s+(NON)?EDITIONABLE)?\\s+(VIEW)\\s+.*\\s+AS\\s+WITH\\s+(FUNCTION|PROCEDURE).*");
```

## Workaround: 

use 5.1.1

pauxus:
Fix #2455:  Replace the regexp of the fix for #2048 with a tighter one

Note that there might still be corner cases where this fails, but this at least fixes most scenarios.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/database/oracle/OracleSqlStatementBuilder.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/database/oracle/OracleSqlStatementBuilder.java
@@ -138,7 +138,7 @@ public class OracleSqlStatementBuilder extends SqlStatementBuilder {
 
     private static final Pattern DECLARE_BEGIN_REGEX = toRegex("DECLARE|BEGIN");
     private static final Pattern PLSQL_VIEW_REGEX = Pattern.compile(
-            "^CREATE(\\s+OR\\s+REPLACE)?(\\s+(NON)?EDITIONABLE)?\\s+(VIEW)\\s+.*\\s+AS\\s+WITH.*");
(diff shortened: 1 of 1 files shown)


MikielAgutu:
It looks like this change is already in the product: https://github.com/flyway/flyway/blob/master/flyway-core/src/main/java/org/flywaydb/core/internal/database/oracle/OracleParser.java#L57

As you said in your [comment ](https://github.com/flyway/flyway/issues/2455#issuecomment-517325801)in #2455:

> This migration works in 6.0.0-beta2, however, we get errors there with valid code, and beta versions are no option for production code.

Am I right in thinking the 6.0.0 release therefore fixes this issue?

Also; 

> we get errors there with valid code

Can you clarify what you meant by this?

pauxus:
Hi, sorry took a while.

I tried our code with the 6.0.0.beta. The error in #2455 did not occur. However we had another (correct) migration that failed. Pinpointing the actual reason is however not easy. 

I did not yet try with 6.0.0.GA, but I assume that the error persists. I assume that this will prevent us from switching to 6, so I think we will create a support request.

Issue #1961, opened 2018-03-23
Support S3 locations


krazavet:
Hello,

I am using Flyway 5.0.7 and I am trying to access to SQL migration scripts located on a S3 repository.
I tried to do so using the configuration file:
```
...
flyway.locations=s3://<bucket>/<path to sql scripts>
...
```
But when running a command, I encountered an error: 
```
$ flyway info                                                                                                                                                                         
Flyway Community Edition 5.0.7 by Boxfuse

ERROR: Unknown prefix for location (should be either filesystem: or classpath:): s3://<bucket>/<path to sql scripts>
```
I am missing something ? 
Or SQL locations are reserved to local filesystem only ?

Thanks in advance

axelfontaine:
Related to #1779

Lyeeedar:
Moving this to flyway 7.x as the v2 SDK requires upgrading flyway from java 7. See https://github.com/flyway/flyway/pull/2458#issuecomment-582872908

MikielAgutu:
@miron4dev This is slated for the V7 release, which we aim to get out summer this year.

sanketh-shetty-incontact:
Hi,

Any update as to when this will be released?

slsshetty:
Hi,

Any updated when will this be released?

RanjithTom:
Any updates on supporting S3 location in flyway.url.locations

DoodleBobBuffPants:
This feature is now available in the V7 beta which you can read more about [here](https://flywaydb.org/blog/flyway-7.0.0-beta).

Issue #1779, opened 2017-09-25
Use java.nio.Path to scan filesystem and read SQL files


gquintana:
This new feature allows:

* To read file from remote file systems: I want to read SQL files from S3 (using [S3](https://github.com/Upplication/Amazon-S3-FileSystem-NIO2) file system provider ) or from a Zip file not in classpath ( [Zip](http://docs.oracle.com/javase/7/docs/technotes/guides/io/fsp/zipfilesystemprovider.html) file system provider) )
* To read file from classloader as if it was as local file: `Path path=Paths.get(classLoader.getResource(resourceName).toURI());`

However, it's probably not compatible with Android as it requires Java 7.

axelfontaine:
Thanks for the suggestion. While those features are neat, the lack thereof hasn't seemed to be much of a deal breaker for users so far. But it's an interesting idea to consider for the future.

gquintana:
Would you accept a PR?

MikielAgutu:
We don't think this is the right thing. S3 locations are on the roadmap : https://github.com/flyway/flyway/issues/1961.

Feel free to create a particular issue for reading from Zip archives etc.

PR #2458, opened 2019-08-04
Support S3 locations

JonLally:
Support s3 locations by use of aws-java-sdk-s3.

Issue references: #1961  #1779 

- I'm not 100% sure of some of the changes I made, specifically the entry in the flyway-core pom for the maven bundle plugin. 
- I also had to explicitly specify 4.5.2 version of the httpclient from apache http components, it is a transitive dependency of aws-java-sdk-s3 and updating the order of the dependencies didn't matter as it took the version from google android always.
- I'm unsure of the testing process, I've had to manually test this but if there is an automated suite I can contribute to I'd like to provide tests around this functionality.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/database/mysql/MySQLParser.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/database/mysql/MySQLParser.java
@@ -113,7 +113,9 @@ public class MySQLParser extends Parser {
     @Override
     protected boolean shouldAdjustBlockDepth(ParserContext context, Token token) {
         TokenType tokenType = token.getType();
-        if (TokenType.DELIMITER == tokenType || ";".equals(token.getText())) {
+        if (TokenType.DELIMITER.equals(tokenType) || ";".equals(token.getText())) {
+            return true;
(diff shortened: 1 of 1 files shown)


KiruthikaSS:
Eager to know, when this feature will be released. Any Milestone planned? Plz let me know.

juliahayward:
Many thanks for the contribution (and welcome to the hall of fame!) We're looking at including this by 6.3 - including tests at our end and documentation changes.

juliahayward:
Do have a look at https://github.com/flyway/flyway/issues/2664 - this came up with both AWS SDK 1.11 and 2.0. We would strongly prefer to use the v2 SDK (which requires some code changes, although they are pretty mechanical and I have done them) however that requires Java 8 - any thoughts?
.

JonLally:
Wasn't aware of the AWS SDK v2, would definitely agree with using it instead of 1.1.

Lyeeedar:
Moving this to flyway 7.x as the v2 SDK requires upgrading flyway from java 7. See https://github.com/flyway/flyway/pull/2458#issuecomment-582872908

Issue #2103, opened 2018-08-09
EsgynDB support


akashshah26595:
###### Which version and edition of Flyway are you using?

- Flyway Community Edition 5.0.7 by Boxfuse (through maven)
- maven dependency:

		<dependency>
			<groupId>org.flywaydb</groupId>
			<artifactId>flyway-core</artifactId>
			<version>5.0.7</version>
		</dependency>

###### If this is not the latest version, can you reproduce the issue with the latest one as well?

- Issue isn't fixed in the latest version (as per release notes)

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin)

- Maven plugin (Trying to run a sample Spring Boot App that connects flyway to Apache Trafodion using its JDBC Driver)

###### Which database are you using (type & version)?

EsgynDB, powered by Apache Trafodion 

###### Which operating system are you using?

Windows 7

###### What did you do?

I'm doing a POC on using flyway with [Esgyn DB](https://esgyn.com/), using trafodion jdbc driver (version jdbcT4-2.4.1). I wrote a script (to create a new table) and placed it under db/migration directory. 
I specified all the database configurations in properties file.

###### What did you expect to see?

When I ran the application, I expected flyway to create two tables (one for schema version and the other one mentioned in the script).

###### What did you see instead?

The following error: 

```
Caused by: org.flywaydb.core.api.FlywayException: Unsupported Database: Trafodion 2.4
	at org.flywaydb.core.internal.database.DatabaseFactory.createDatabase(DatabaseFactory.java:179) ~[flyway-core-5.0.7.jar:na]
```
I traced back the error to DatabaseFactory class where a fixed set of supported databases are specified.
It doesn't have trafodion or any other database that can be accessed using generic jdbc drivers.

#### Is there a way to use flyway with unsupported databases (using database specific drivers)? 

Do let me know if any additional information is required from me.

eowhadi:
+1, I would like to add support of Apache Trafodion/EsgynDB to Apache fineract cn, but fineract cn uses flyway. So support for this feature is a pre-requisite.

RohitJain13:
We are almost ready to add Apache Trafodion and EsgynDB support.  We have tested it and are looking to get a few questions resolved before doing so.

DoodleBobBuffPants:
Duplicate of https://github.com/flyway/flyway/issues/2451

PR #2476, opened 2019-08-20
EsgynDB and Apache Trafodion Support

RohitJain13:
Added support for EsgynDB and Apache Trafodion to address issue EsgynDB support #2103.  JDBC drivers for both would be needed for final product deployment.


piers-williams:

PR #2509, opened 2019-09-19
Feature #1737 - intermediate baselines

### LINKED ISSUE #470: Condense existing migrations into a new initial one
_Original author: theasp...@gmail.com (March 29, 2013 11:29:05)_

We have hundred migrations, tabels created and drops, unnecessary fields and so on. It will be great to add cummulative migration which can be applied to clean schema to optimize initial migration.

_Original issue: http://code.google.com/p/flyway/issues/detail?id=470_

### LINKED ISSUE #1737: Re-BaseLine and Initial Script
I wonder if there are 2 features that already in planning / existing / similar with those 2 that I need:

# Re-Baseline
No, I don't talk about feature that has been mentioned in #908.

Re-baseline simply add a baseline record in schema version table. For example:

| version | script |
| --- | --- |
| 1.0.0 | << Flyway Baseline >> |
| 1.1.0 | V1.1.0__migrate.sql |
| 1.2.0 | V1.2.0__migrate.sql |
| 1.2.3 | V1.2.3__bugfix.sql |
| 2.0.0 | << Flyway Baseline >> |

The purpose is only to prevent flyway migrate to execute the migration version 2.0.0. Sql script V2.0.0 is a script that construct database from scratch, up to latest migration or V1.2.3 in this case. The script will be similar with what will be produced if issue #470 is delivered, but manually constructed by developer.

# Initial Script
Continuing from previous example, I want the initial flyway migrate to start from script V2.0.0. without executing the artifacts before V2.0.0.

konopka:
This feature allows one to add intermediate baselines while keeping all other migrations in place.
Default naming convention is using 'IB' prefix => 'IB3__squashed_migrations_1_upto_3'
The logic is as follows:
1. when there are any migrations applied, then intermediate baseline migrations will be ignored
2. when no migrations were applied yet, then latest intermediate baseline migration will be picked as 'baseline'

Resolves #1737, resolves #470


alextercete:
@konopka Thanks for your contribution, it looks very promising!

We have to consider the impact of this change and how it fits with the existing baselining mechanism, so it might a little while before it's incorporated and released.

konopka:
@alextercete Thanks for the info. I am glad you like it. 
If there is anything you want me to change I am willing to update this PR
I would really appreciate (and I think many others as well) if this makes it in.

donbeave:
Looks very useful feature, now we have almost 1500 migrations, initializing then takes a lot of time, it will be nice to combine all of them into one file, not just concat, but dump the state after applying them. If I understand correctly this PR is all about same issue. Is there any estimates when it will be merged?

MikielAgutu:
I believe problem can be solved with the new [`skipExecutingMigrations`](https://flywaydb.org/documentation/configuration/skipExecutingMigrations) option in V7. You can do the following:
- Set [`target`](https://flywaydb.org/documentation/configuration/target) to your new 'baseline' (e.g. `target=1500`)
- Run migrate with `skipExecutingMigrations=true`
- All migrations up to and including V1500 are recorded as being executed in the schema history, but aren't actually executed

Does that help?

konopka:
Hi @MikielAgutu 
I do not think the skipExecutionMigration feature is a good thing for production environments.
1. It requires manual interaction, and whole point of having flyway is to automate things on deploys
2. You are losing visibility of what was executed and when, history table will become a blob of fake data

jaredmdobson:
@DoodleBobBuffPants any chance we can get this added in as a feature?

konopka:
@DoodleBobBuffPants  @MikielAgutu checking again if there is any chance of this getting in?
On my current project we are rapidly closing in on 1000 migrations mark, which is a problem for many others I believe.

MikielAgutu:
To give an update on this - we're still considering how best to solve this problem. I think this approach has merit, but it's not necessarily the direction we want to take with Flyway.

If I understand correctly, an 'intermediate baseline' can be likened to a snapshot of the database. It represents the state of the DB at a given point in time. Generally it's the resultant state after a set of migrations - but it could be something arbitrary too, in principle at least. The latest IB/snapshot is the true 'baseline'.

In a sense, the baseline can therefore 'move' upwards as time goes on. Executing the latest baseline is more performant for new deployments compared to executing many migrations, which may have contradictory changes.

e.g. V1 creates table, V5 deletes it, - so why not start from the resultant state of the DB at V5?

Older baselines are disregarded - although preserved in version control presumably because they have some value for audit purposes.

Is that understanding correct?

If so, then I agree this is a problem that needs solving in Flyway. However we need time to consider how to solve this problem in a coherent way.

For instance, what becomes of the existing baseline command? We might need to reform how it works in order for the concept of an 'intermediate' baseline to make sense. And I'm sure there are other impacts this work would have that we don't quite have scope to think about right now.

So we can't commit to merging this right away - it isn't at the top of our priority list. However we will re-consider it for Flyway V8.

konopka:
Thanks @MikielAgutu for the update.
Yes, your understanding of the problem is correct. This comment sums it up very nicely: https://github.com/flyway/flyway/issues/470#issuecomment-243387349

mcekovic:
This is definitely a very useful feature, but unfortunately it does not solve #470 in full.
It helps when brand new environment is to be installed when there are a lot of migrations.
However, in product development, where there are a lot of production environments at different customers, and in case a couple of customers start to lag in applying migrations for a long time, upgrading such customers to the mainline with a lot of migrations is still unsolved.
Say there are 1000 migrations and say there is an intermediate baseline every 100 migrations.
If we have an environment that is on migration 350 and has to be upgraded to migration 1000, we have to apply 650 migrations, which may be very time consuming, leading to sometimes unacceptable production downtime.
Am I missing something in the proposed feature, can it solve this issue?

DoodleBobBuffPants:
We now support this with [state scripts](https://flywaydb.org/documentation/concepts/statescripts) so closing this PR

Do let us know if this works as intended :)

konopka:
@DoodleBobBuffPants  You took public contribution [with obvious minor changes] and made it available only in paid version of the product.
Not very nice of you, not at all.
I guess then, it is what it is, and I'll have to continue using custom builds indefinitely :-(

seanf:
For anyone else finding this old discussion, apparently [state scripts](https://web.archive.org/web/20210919141624/https://flywaydb.org/documentation/concepts/statescripts) are now known as [**baseline migrations**](https://documentation.red-gate.com/flyway/flyway-concepts/migrations/baseline-migrations), and it looks like they [no longer require Flyway Teams](https://documentation.red-gate.com/flyway/learn-more-about-flyway/feature-summary).

PR #2554, opened 2019-11-05
Gradle Plugin: Fix reading CONFIG_FILES from project properties

kekru:
Hello,
thanks for great plugin :)

I have found a small issue in the Gradle Plugin:
When reading CONFIG_FILES from getProject().getProperties(), it was wrongly read from System.getProperties()

Greetings
Kevin


alextercete:
@kekru Good catch, thanks for your contribution!

Issue #2626, opened 2020-01-15
Support for MariaDB enterprise version numbering


NicholasFechner:
###### Which version and edition of Flyway are you using?
Flyway 6.0.8 (linked by Spring Boot 2.2.2)

###### If this is not the latest version, can you reproduce the issue with the latest one as well?
(Many bugs are fixed in newer releases and upgrading will often resolve the issue)
Yes.

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin)
org.mariadb.jdbc:mariadb-java-client:2.4.4 (also via Spring Boot)

###### Which database are you using (type & version)?
MariaDB 10.4 Enterprise version.

###### Which operating system are you using?
Red Hat Enterprise Linux 7

###### What did you do?

Starting the application with Flyway active runs into an error:

```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'flywayInitializer' defined in class path resource [org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration$FlywayConfiguration.class]: Invocation of init method failed; nested exception is org.flywaydb.core.api.FlywayException: Unable to determine version from '10.4.10-4-MariaDB-log'
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.initializeBean(AbstractAutowireCapableBeanFactory.java:1796)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:595)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:517)
	at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:323)
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:222)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:321)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:202)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:310)
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:202)
[...]

PR #2627, opened 2020-01-15
#2626 Support for MariaDB enterprise version numbering

NicholasFechner:
Simple fix for the different versioning scheme of MariaDB enterprise versions.


juliahayward:
Great, thankyou!

PR #2635, opened 2020-01-20
Fix #2634 deadlock in Postgresql 12

### LINKED ISSUE #2634: Deadlock in PostgreSQL 12
Flyway version: **6.1.4 (API)**
DB version: **PostgreSQL 12.1 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 9.2.0, 64-bit**

This code is calling flyway on start up:
```java
Flyway flyway = Flyway.configure()
        .baselineOnMigrate(true)
        .dataSource(dataSource())
        .load();
flyway.migrate();
```
Then flyway tries to lock table, but it fails on this line
https://github.com/flyway/flyway/blob/72cfe7db4a21a44001648ea8e00aef70849cf24d/flyway-core/src/main/java/org/flywaydb/core/internal/database/postgresql/PostgreSQLAdvisoryLockTemplate.java#L112
because `rs.getString("pg_try_advisory_lock")` returns `true`, so we've got infinite loop

romash1408:


flyway-core/src/main/java/org/flywaydb/core/internal/database/postgresql/PostgreSQLAdvisoryLockTemplate.java
```diff
@@ -109,7 +109,7 @@ private boolean tryLock() throws SQLException {
                 new RowMapper<Boolean>() {
                     @Override
                     public Boolean mapRow(ResultSet rs) throws SQLException {
-                        return "t".equals(rs.getString("pg_try_advisory_lock"));
+                        return rs.getBoolean(0);
```
juliahayward: Column indexes are 1-based - otherwise this is a definite improvement!

alextercete: Couldn't we stick to using the column name instead of the index?

PR #2637, opened 2020-01-21
Add environment variable for defaultSchema

rickselby:
There is currently no way to define the default schema through environment variables.


alextercete:
@rickselby Good spot, thanks!

Issue #1685, opened 2017-06-29
Incorrect Table implementation for Vertica


avmukilan:
###### What version of Flyway are you using?

Flyway 4.2.0

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin, SBT plugin, ANT tasks)

Command line

###### What database are you using (type & version)?

Vertica 

###### What operating system are you using?

linux

###### What did you do?
(Please include the content causing the issue, any relevant configuration settings, and the command you ran)

What to create a new management system using flyway

###### What did you expect to see?

Unable to check whether table "***_test"."schema_version" exists
----------------------------------------------------------------
SQL State  : 3F000
Error Code : 4650
Message    : [Vertica][VJDBC](4650) ERROR: Schema "information_schema" does not exist

It fires a wrong query to check for the table "***_test"."schema_version" 

###### What did you see instead?
It should ideally create a schema_version under ***_test ( name masked by *** )

and have a correct implementation for ( VerticaSchema.java)
@Override
    public Table getTable(String tableName) {
        return new PostgreSQLTable(jdbcTemplate, dbSupport, this, tableName);
    }

nop-s:
This patch will correct the query issue for Vertica.

[vertica_support_queries.patch.gz](https://github.com/flyway/flyway/files/1539953/vertica_support_queries.patch.gz)

axelfontaine:
@nop-s Thanks. We'll be revisiting Vertica support in detail for 5.1.

alattkechewy:
@axelfontaine we are considering Flyway for use with Vertica 8.1. Do you know if support will be re-introduced in version 6.0.0? Thank you for all the great work!

MikielAgutu:
See https://github.com/flyway/flyway/issues/1761#issuecomment-580676739

PR #2651, opened 2020-01-30
Applying patch to make Vertica work

tomdubiel:
- vertica doesnt use the information_schema table, it uses v_catalog
- flyway 4.2.0 pre-patch was using PSQL as a template for vertica
- the patch provides vertica-specific handling
- details: https://github.com/flyway/flyway/issues/1685


tomdubiel:
woooops, this should have been against a fork

PR #2670, opened 2020-02-11
Custom delimiter

JFNGO:
The goal here is to allow to set a custom delimiter using parameters in the conf file.
Some examples and tests on the new feature can be found in the following file:
[TestCaseFlyway.docx](https://github.com/flyway/flyway/files/4188874/TestCaseFlyway.docx)


MikielAgutu:
Related to https://github.com/flyway/flyway/pull/2670

MikielAgutu:
We're not sure about these changes. It looks like you're trying to add a Flyway feature that clashes with existing SQL language features.

Flyway should support any valid SQL. So if you can run it in SQL client for your DB engine, it should work. We consider it a bug when this isn't the case.

We don't want to add a generic way of changing delimiters when some DB engines already support it in their own way. Instead we should ensure we support the language features of the SQL flavors we support.

For instance, have you tried the various options for Oracle delimiter configuration? Like [`SET BLOCKTERMINATOR`](https://docs.oracle.com/cd/E11882_01/server.112/e16604/ch_twelve040.htm#SQPUG067) etc. It should work with Flyway. If it doesn't that's a bug that should be fixed as a localized change in the context of the existing Oracle parsing.

JFNGO:
Thank you for your email.
We will take that into consideration and work on the Oracle Parser to allow
the use of '/' as delimiter on a new line even for standard statements.

Le mer. 12 févr. 2020 à 08:46, Mikiel Agutu <notifications@github.com> a
écrit :

PR #2745, opened 2020-03-27
Spanner support with Open Source JDBC driver

stankiewicz:
Hi,
Due to open sourcing Spanner driver and it's better support for DML and DDL it is possible to integrate it with flyway. 
There are still things not supported like transactional ddl/dml.
Also due to the fact that information schema needs read only connection I had to use raw connection instead of jdbcTemplate. 

please take a look.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,11 +80,11 @@
         <snapshotRepository>
             <id>${snapshotRepository.id}</id>
             <name>${snapshotRepository.name}</name>
-            <url>${releaseRepository.url}</url>
+            <url>${snapshotRepository.url}</url>
         </snapshotRepository>
         <repository>
-            <id>${repository.id}</id>
-            <name>${repository.name}</name>
+            <id>${releaseRepository.id}</id>
+            <name>${releaseRepository.name}</name>
             <url>${releaseRepository.url}</url>
         </repository>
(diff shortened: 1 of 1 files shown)


stankiewicz:
this PR should replace #1880  done by @olavloite . #1880 was done using unofficial driver.

thinhha:
+1 this is very much needed.

stankiewicz:
@MikielAgutu , would you be able to take a look at this?

MikielAgutu:
@stankiewicz Thanks for the contribution! The new JDBC driver certainly provides more confidence this would be a stable addition to Flyway.

Because this is a cloud database, it's more difficult for us to set up automated testing + CI processes to maintain a high level of stability. We'll need some time to decide internally if this is the right direction for Flyway.

We acknowledge this is a popular DB and we are pleased that the community has done so much work to contribute support for it.

stankiewicz:
Hi @MikielAgutu , thank you for putting it on radar - Spanner have emulator. 
If you have requirements somewhere how it is tested then I'm more then happy to contribute in this area.

stankiewicz:
@MikielAgutu , I have created new PR (#2776) due to the fact that this PR had very old changes, binary (jar) commit and it was impossible to squash..

Issue #2123, opened 2018-08-29
Performance issue for the MySQLSchema doEmpty() call


Val870:
###### Which version and edition of Flyway are you using?
5.0.7. 
But  it's irrelevant as it's still in [the last version of the file](https://github.com/flyway/flyway/blob/master/flyway-core/src/main/java/org/flywaydb/core/internal/database/mysql/MySQLSchema.java)

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin)
Java via Maven plugin

###### Which database are you using (type & version)?
MySQL 5.6 and 5.7

###### Which operating system are you using?
Linux. I don't put more info on that because it's (a priori) not relevant here.

###### Usage context
We have 1086 schema. In these schema, ~1000 have the same structure with 845 tables and views.
So our `information_schema` is a little heavy. 

###### What did you do?
`flyway.migrate()` on a non Flyway migrated yet db (and so with `baselineOnMigrate` to true).

###### What did you expect to see?
Relatively fast first migration.

###### What did you see instead?
Relatively slow first migration.

###### Why ?
The db has not been migrated with Flyway yet. So `schemaHistory` is empty.
And, by following the Flyway process from [here](https://github.com/flyway/flyway/blob/master/flyway-core/src/main/java/org/flywaydb/core/Flyway.java#L1232), a `schema.empty()` is called, which call the `doEmpty()` of `MySQLSchema` (in our case). 
But the `doEmpty()` is done via the `information_schema`. And takes approximately 1min to do this method in our case. 

###### So ? What is the question ?
Why did you not use the [SHOW methods](https://dev.mysql.com/doc/refman/8.0/en/show.html) ?
They are much more efficient than go through the `information_schema`.
And another question would be is this performance problem can also exist on the other `Schema` children ?

Thanks for the reading and have a great day !

axelfontaine:
I don't believe using `SHOW` will improve performance as that will most likely use the `information_schema` behind the scenes anyway.

What definitely should be done though is replacing those `SELECT COUNT(*)` with either `EXISTS` or `SELECT 1 ... LIMIT 1`.

Val870:
Seems legit. 

If you wish, I can try the query beforehand. 
What takes time on the current query is the `Select count(*) from information_schema.TRIGGERS Where TRIGGER_SCHEMA=?`

axelfontaine:
Yes, could you try:

```sql
Select 
[...]

PR #2750, opened 2020-03-31
Fix slow doEmpty by using EVENT_OBJECT_SCHEMA (#2123)

AlexisDeschamps:
Associated issue: #2123

This fix improves the performance of `MySQLSchema.doEmpty` for MySQL 5.7 without changing the expected behavior.

The subquery `SELECT 1 as found FROM information_schema.triggers WHERE trigger_schema=? AND  LIMIT 1` is changed to `SELECT 1 as found FROM information_schema.triggers WHERE event_object_schema=? AND  LIMIT 1` (`WHERE trigger_schema=?` becomes `WHERE event_object_schema=?`).

As specified here: https://dev.mysql.com/doc/mysql-infoschema-excerpt/5.7/en/triggers-table.html, "every trigger is associated with exactly one table" and `EVENT_OBJECT_SCHEMA` is the "schema (database) in which this table occurs" and `TRIGGER_SCHEMA` is the "name of the schema (database) to which the trigger belongs". If a trigger belongs to a table that belongs to schema X, then that trigger must belong to schema X. Thus, if one's goal is to find if there are any triggers associated with a specific schema, one can filter by `EVENT_OBJECT_SCHEMA`.

As specified here: https://dev.mysql.com/doc/refman/5.7/en/information-schema-optimization.html, filtering by `EVENT_OBJECT_SCHEMA` is faster than filtering by `TRIGGER_SCHEMA` because it is a "Column to specify to avoid data directory scan".


juliahayward:
Many thanks for the contribution!

juliahayward:
Just to confirm, this should be included in the 6.4 release as well.

PR #2776, opened 2020-04-20
Spanner support with Open Source JDBC driver

stankiewicz:
Hi,
Due to open sourcing Spanner driver and it's better support for DML and DDL it is possible to integrate it with flyway.
There are still things not supported like transactional ddl/dml.
Also due to the fact that information schema needs read only connection I had to use raw connection instead of jdbcTemplate.

please take a look.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/database/cloudspanner/CloudSpannerParser.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/database/cloudspanner/CloudSpannerParser.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2010-2020 Redgate Software Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.flywaydb.core.internal.database.cloudspanner;
+
+import java.util.List;
+import org.flywaydb.core.api.configuration.Configuration;
+import org.flywaydb.core.api.logging.Log;
+import org.flywaydb.core.api.logging.LogFactory;
+import org.flywaydb.core.internal.parser.Parser;
+import org.flywaydb.core.internal.parser.ParsingContext;
+import org.flywaydb.core.internal.parser.Token;
+
+public class CloudSpannerParser  extends Parser {
+    private static final Log LOG = LogFactory.getLog(CloudSpannerParser.class);
+    public CloudSpannerParser(Configuration configuration, ParsingContext parsingContext) {
+        super(configuration,parsingContext, 3);
+    }
+
+    @Override
+    protected char getIdentifierQuote() {
+        return '`';
+    }
+
+    protected char getAlternativeIdentifierQuote() {
+        return '\"';
+    }
+
+    @Override
+    protected Boolean detectCanExecuteInTransaction(String simplifiedStatement,
+                                                    List<Token> keywords) {
+        LOG.debug("checking if [" + simplifiedStatement + "] can run in transaction");
+        // Flyway tries to do hold transaction in which migration will happen
+        return false;
+    }
+}
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/database/DatabaseFactory.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/database/DatabaseFactory.java
@@ -21,6 +21,8 @@ import org.flywaydb.core.api.logging.Log;
 import org.flywaydb.core.api.logging.LogFactory;
 import org.flywaydb.core.internal.callback.CallbackExecutor;
 import org.flywaydb.core.internal.database.base.Database;
+import org.flywaydb.core.internal.database.cloudspanner.CloudSpannerDatabase;
+import org.flywaydb.core.internal.database.cloudspanner.CloudSpannerParser;
 import org.flywaydb.core.internal.database.cockroachdb.CockroachDBDatabase;
 import org.flywaydb.core.internal.database.cockroachdb.CockroachDBParser;
 import org.flywaydb.core.internal.database.cockroachdb.CockroachDBRetryingStrategy;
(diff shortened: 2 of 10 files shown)


stankiewicz:
@MikielAgutu, this PR is new version of old #2745 . It's one commit, based on latest master so it should easy to merge.

stankiewicz:
also, I'm happy to help with setting up testing with spanner emulator, just let me know how I can contribute.

MikielAgutu:
To add more detail to https://github.com/flyway/flyway/pull/2745#issuecomment-611534096:

We've looked at the emulator, and it seems promising. However it does have some [limitations](https://github.com/GoogleCloudPlatform/cloud-spanner-emulator/blob/master/README.md#features-and-limitations). It's also in beta apparently.

The limitations mean I'm not sure it'll give us the confidence we need to assure a good level of support in Flyway. We'll continue to monitor it, and review when the emulator exits beta.

stankiewicz:
@MikielAgutu , agree on limitations but we've given a try to it in other schema management solution and it could cover the scenarios. It's in beta but it will exit it soon (can't disclose dates). Any way I could help you with setting up testing or just check test scanarios?

MikielAgutu:
@stankiewicz Thanks for your offer of help! When/if we commit to this work we may get in contact.

Lyeeedar:
@stankiewicz We tried to test this PR using the open source driver, but it locks forever when trying to execute any queries deep inside the driver itself.

```park:-1, Unsafe (jdk.internal.misc)
parkNanos:234, LockSupport (java.util.concurrent.locks)
awaitFulfill:462, SynchronousQueue$TransferStack (java.util.concurrent)
transfer:361, SynchronousQueue$TransferStack (java.util.concurrent)
poll:937, SynchronousQueue (java.util.concurrent)
pollUninterruptiblyWithTimeout:919, SessionPool$Waiter (com.google.cloud.spanner)
take:893, SessionPool$Waiter (com.google.cloud.spanner)
access$4600:876, SessionPool$Waiter (com.google.cloud.spanner)
getReadSession:1390, SessionPool (com.google.cloud.spanner)
getReadSession:55, DatabaseClientImpl (com.google.cloud.spanner)
singleUseReadOnlyTransaction:140, DatabaseClientImpl (com.google.cloud.spanner)
executeQuery:172, SingleUseTransaction (com.google.cloud.spanner.jdbc)
internalExecuteQuery:777, ConnectionImpl (com.google.cloud.spanner.jdbc)
parseAndExecuteQuery:708, ConnectionImpl (com.google.cloud.spanner.jdbc)
executeQuery:681, ConnectionImpl (com.google.cloud.spanner.jdbc)
executeQuery:163, AbstractJdbcStatement (com.google.cloud.spanner.jdbc)
executeQueryWithOptions:61, JdbcPreparedStatement (com.google.cloud.spanner.jdbc)
getTables:761, JdbcDatabaseMetaData (com.google.cloud.spanner.jdbc)
doExists:48, CloudSpannerTable (org.flywaydb.core.internal.database.cloudspanner)
exists:58, Table (org.flywaydb.core.internal.database.base)
exists:92, JdbcTableSchemaHistory (org.flywaydb.core.internal.schemahistory)
allAppliedMigrations:181, JdbcTableSchemaHistory (org.flywaydb.core.internal.schemahistory)
refresh:131, MigrationInfoServiceImpl (org.flywaydb.core.internal.info)
info:52, DbInfo (org.flywaydb.core.internal.command)
execute:338, Flyway$5 (org.flywaydb.core)
execute:331, Flyway$5 (org.flywaydb.core)
execute:527, Flyway (org.flywaydb.core)
info:331, Flyway (org.flywaydb.core)
executeOperation:204, Main (org.flywaydb.commandline)
main:127, Main (org.flywaydb.commandline)```

Looking at your branch it looks like you only had a demo for the non-open source driver?

**Comment by skuruppu** [comment]
I am CC'ing @olavloite here who is the developer for the Cloud Spanner JDBC driver.

@olavloite, would you be able to take a look at why the JDBC driver may lock when trying to query a Cloud Spanner database?

**Comment by stankiewicz** [comment]
Thanks @skuruppu for cc'ing @olavloite . 
com.google.cloud:google-cloud-spanner-jdbc is open source driver. 
What was your setup - are you using service account or user credentials or maybe are you using emulator?

I've checked this with my simple demo project - https://github.com/stankiewicz/flywaydemo and it works for regular spanner database with service account setup.

**Comment by olavloite** [comment]
@Lyeeedar I have only experienced that the driver hangs like that indefinitely when trying to connect to a non-existent or invalid host. The most common case for that is trying to connect to the emulator while using SSL. The emulator only supports plain text communication. If you want to connect to the emulator, you should use one of the following options:
1. Set the environment variable `SPANNER_EMULATOR_HOST` to the gRPC host name and port of the emulator (mostly `localhost:9010`). The connection URL should in that case not contain any host name or other special properties. The JDBC driver and client library will handle everything for you.
1. OR include the host name and a property telling the driver to use plain text communication in the connection URL. That would look like this: `jdbc:cloudspanner://localhost:9010/projects/my-project/instances/my-instance/database/my-databases;usePlainText=true`.

**Comment by Lyeeedar** [comment]
> @Lyeeedar I have only experienced that the driver hangs like that indefinitely when trying to connect to a non-existent or invalid host. 

Is there a way we can make the driver fail faster. As it currently seems to hang indefinitely, it makes it look like Flyway has deadlocked, instead of informing the user what the actual issue is.

**Comment by olavloite** [comment]
> > @Lyeeedar I have only experienced that the driver hangs like that indefinitely when trying to connect to a non-existent or invalid host.
> 
> Is there a way we can make the driver fail faster? As it currently seems to hang indefinitely, it makes it look like Flyway has deadlocked, instead of informing the user what the actual issue is.

I'll have a look and see if that is doable. While it sounds trivial, it is a little bit tricky as this happens on the network level for the first actual server request that is being done, and the first request will differ depending on both the further configuration of the connection, as well as the first statement the user issues. But it might be worthwhile adding some kind of ping statement to check that a valid connection can be made directly after opening the connection, but before returning it to the user.

File: flyway-core/src/main/java/org/flywaydb/core/internal/database/cloudspanner/CloudSpannerParser.java [hunk 1]
```diff
@@ -0,0 +1,48 @@
...
+        super(configuration,parsingContext, 3);
+    }
+
+    @Override
+    protected char getIdentifierQuote() {
+        return '`';
+    }
+
+    protected char getAlternativeIdentifierQuote() {
+        return '\"';
```
olavloite, line 38: Only backticks are allowed for quoting identifiers. Enclosing a literal or identifier in double quotes will make Spanner interpret it as a string literal.

DoodleBobBuffPants:
Thanks for the PR! This is now available in the V7 beta which you can read more about [here](https://flywaydb.org/blog/flyway-7.0.0-beta).

PR #2780, opened 2020-04-21
Add jdbc-secretsmanager drivers to DriverDataSource

PetterKraabol:
We use Amazon's [jdbc-secretsmanager](https://github.com/aws/aws-secretsmanager-jdbc) for database authentication with secrets from Secrets Manager.

As pointed out by https://github.com/aws/aws-secretsmanager-jdbc/issues/23, Amazon's library unfortunately uses a prefix `jdbc-seceretsmanager:` which makes Flyway throw here https://github.com/flyway/flyway/blob/1e2be7a211f1ed8d6176f95d4b83b0f5dd69f01b/flyway-core/src/main/java/org/flywaydb/core/internal/jdbc/DriverDataSource.java#L215-L217

This PR adds available prefixes and corresponding drivers to Flyway's [DriverDataSource](https://github.com/flyway/flyway/blob/1e2be7a211f1ed8d6176f95d4b83b0f5dd69f01b/flyway-core/src/main/java/org/flywaydb/core/internal/jdbc/DriverDataSource.java) to handle jdbc-secretsmanager urls.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/jdbc/DriverDataSource.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/jdbc/DriverDataSource.java
@@ -54,16 +54,21 @@ public class DriverDataSource implements DataSource {
         INFORMIX("jdbc:informix-sqli:", "com.informix.jdbc.IfxDriver"),
         JTDS("jdbc:jtds:", "net.sourceforge.jtds.jdbc.Driver"),
         MARIADB("jdbc:mariadb:", "org.mariadb.jdbc.Driver"),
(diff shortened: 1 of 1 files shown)


MikielAgutu:
We'd prefer if Amazon would just update their driver to use the more common `jdbc` prefix. The driver should play nice with existing conventions. We don't really want to maintain this knowledge within the Flyway codebase.

Lyeeedar:
Thank you for the contribution. Due to an internal refactor of the database code for Flyway v7, we won't be accepting this PR. However we have reimplemented this in our own code so from Flyway v7 onwards you will be able to use `jdbc-secretsmanager:` urls.

PR #2790, opened 2020-04-27
maven-jar-plugin 3.2.0

sullis:


MikielAgutu:
Thanks for the contribution. Is there a particular reason why you would like this dependency updated?

juliahayward:
(Bumped to 3.2.0 in the main repo, pushed across - same effect as this PR)

PR #2793, opened 2020-04-28
#2698. Configure custom comparator defining repeatable migrations order.

### LINKED ISSUE #2698: Define script execution order
Introduce a mechanism to specify the order of migrations beyond filenames. Things like interleaving repeatable and versioned execution, etc.

Related issues:
https://github.com/flyway/flyway/issues/2541
https://github.com/flyway/flyway/issues/2610
https://github.com/flyway/flyway/issues/2647
https://github.com/flyway/flyway/issues/2683
https://github.com/flyway/flyway/issues/1858

antonbakalets:
I would like to control the order of repeatable migrations. 
In order to do that I suggest a new configuration parameter `repeatableMigrationComparator`, fully qualified class name of custom Comparator to determine the order of repeatable migrations.

By default repeatable migrations are ordered by description. Which means as before if not set.

This doesn't brake the main idea that versioned migrations are always applied before repeatable. The custom Comparator implementation controls only repeatable migration order.

I have found that it might close #2698 .

### DIFF
--- a/flyway-commandline/src/main/assembly/flyway.conf
+++ b/flyway-commandline/src/main/assembly/flyway.conf
@@ -109,6 +109,10 @@
 # defined by 'flyway.resolvers' are used. (default: false)
 # flyway.skipDefaultResolvers=
 
+# Fully qualified class name of custom Comparator to determine the order of repeatable migrations
+# By default repeatable migrations are ordered by description.
+# flyway.repeatableMigrationComparator=
+
 # Comma-separated list of directories containing JDBC drivers and Java-based migrations.
 # (default: <INSTALL-DIR>/jars)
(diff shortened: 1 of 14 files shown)


MikielAgutu:
Thanks for the contribution. We are planning to do something like this, however the change would be quite a fundamental extension to the way Flyway works. We'd like to solve the problem in a more general way - for instance, to help with versioned migrations too (e.g. interleaving with repeatables). 

There's a fair bit we need to consider, so we're holding off until V7 at the earliest. We'll keep this PR open as a reference.

DoodleBobBuffPants:
Closing as we will be tackling the problem of custom migration ordering differently

PR #2808, opened 2020-05-05
Placeholder names set by environment variables are case sensitive.

radzimir:
We use uppercase placeholders like ${TABLESPACE} set in flyway.conf as
flyway.placeholders.TABLESPACE=USERS

then I'v tried to set 

export FLYWAY_PLACEHOLDERS_TABLESPACE=USERS

but it hasn't worked because substring after FLYWAY_PLACEHOLDERS_ was lowercased.

This pull request makes setting uppercased placeholders possible, but could potencially breake some projects that set it as uppercase but use as lovercase.

So I'm not sure if my fix is correct one.

Alternativ, case-insensitivity could be added while using placeholders, then it would be irrelevant how we set them.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/configuration/ConfigUtils.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/configuration/ConfigUtils.java
@@ -113,7 +113,7 @@ public class ConfigUtils {
             String convertedKey = convertKey(entry.getKey());
             if (convertedKey != null) {
                 // Known environment variable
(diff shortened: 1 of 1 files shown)


MikielAgutu:
Good spot. You're right this is going to be a breaking change. We can consider it for version 7.

MikielAgutu:
@radzimir Could you sign the CLA?

radzimir:
> Could you sign the CLA?

Sorry, I made a commit with wrong email adress and can not singn CLA. I need to redo it again with a correct one.

radzimir:
I must do pull request again in correct way.

PR #2814, opened 2020-05-14
Memory wise sql migration

mitch2na:
I had an issue with a big sql migration consuming the memory on the running machine. This fixed the issue for me so figured i would submit a PR with a suggestion on how to address the issue

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/sqlscript/ParserSqlScript.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/sqlscript/ParserSqlScript.java
@@ -82,13 +82,14 @@ public class ParserSqlScript implements SqlScript {
                 SqlStatement sqlStatement = sqlStatementIterator.next();
                 if (inMemory) {
                     long freeMemory = getFreeMemory();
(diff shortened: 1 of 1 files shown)


MikielAgutu:
Thank you for the contribution. Can you explain in a bit more detail the problem you experienced, and how these changes fix that problem?

mitch2na:
Yea for sure, the issue occurs when the machine you are running flway from, has no memory left to process the sql migration script, since each migration loads all its statements into an array before executing.

For example, i have a 1 gb server that needed to run a 5 gb sql migration script. Eventually i ran out of memory and the server hangs.

This PR monitors the memory while it's loading sql statements into the array, if it notices the machines free memory is less than 300,000 bytes or 0.03 mb it switches to changing SqlStatementIterator to be a buffered file reader.

MikielAgutu:
This should already be solved with the [`stream` configuration option](https://flywaydb.org/documentation/commandline/migrate#stream). It's for Flyway Pro & Enterprise customers only.

mitch2na:
Ahh gotcha! Okay ill just have to use my fork moving forward for my project. Thanks for the pointing me to the docs

PR #2882, opened 2020-07-16
Placeholder names set by environment variables made case sensitive.

radzimir:
NOTE: this is a second try of pull request after problem with CLA

We use uppercase placeholders like ${TABLESPACE} set in flyway.conf as
flyway.placeholders.TABLESPACE=USERS

then I'v tried to set

export FLYWAY_PLACEHOLDERS_TABLESPACE=USERS

but it hasn't worked because substring after FLYWAY_PLACEHOLDERS_ was lowercased.

This pull request makes setting uppercased placeholders possible, but could potencially breake some projects that set it as uppercase but use as lovercase.

So I'm not sure if my fix is correct one.

Alternativ, case-insensitivity could be added while using placeholders, then it would be irrelevant how we set them.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/configuration/ConfigUtils.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/configuration/ConfigUtils.java
@@ -113,7 +113,7 @@ public class ConfigUtils {
             String convertedKey = convertKey(entry.getKey());
             if (convertedKey != null) {
                 // Known environment variable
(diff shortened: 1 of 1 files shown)


MikielAgutu:
Having thought about this more we aren't sure this is the right thing for Flyway. We think a better solution would be to make environment variable placeholder names case insensitive

radzimir:
OK, I'm closing this pull request.

PR #2901, opened 2020-08-11
Perform disk operation in right side of or statement

ekennedy247:
Hey, realize this isn't fully inline with the contributing guidelines, but the fix is pretty small so I opened a PR without an issue. If that's an issue (haha) I can re-open this.

I noticed with 6.5.3 that an application I have that has a few hundred migrations now takes ~30 seconds (used to be ~2 seconds on 6.5.2) to run in my CI environment (it's an old amazon linux ami, if it matters). Locally (on OSX), performance seems a little worse (~5 seconds -> ~8 seconds). It seems like it's because of commit https://github.com/flyway/flyway/commit/ecfebd2e99f9ac507c3af3627dad6d149dc89335. 

This PR inlines the `getAbsolutePathOnDisk` call to the right side of the `or` statement so it can be skipped when the cheap `getRelativePath` equality check is true.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/scanner/Scanner.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/scanner/Scanner.java
@@ -74,9 +74,8 @@ public class Scanner<I> implements ResourceProvider, ClassProvider<I> {
     @Override
     public LoadableResource getResource(String name) {
         for (LoadableResource resource : resources) {
-            String absolutePath = resource.getAbsolutePathOnDisk();
             String relativePath = resource.getRelativePath();
-            if (relativePath.equals(name) || absolutePath.equals(name)) {
+            if (relativePath.equals(name) || resource.getAbsolutePathOnDisk().equals(name)) {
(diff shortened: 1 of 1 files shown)


DoodleBobBuffPants:
Thank you for your submission! We will look into merging this soon :)

kalejami:
I have a question on this issue. We are using Windows in our company and since 6.5.3 our migrations needed about 20 to 30 times more time to run. 

On the first view I thought that this MR would fix this problem, but it didn't. I debugged the code a little bit and realized that the method `public LoadableResource getResource(String name)` is only called for the "Script Config Files"-Feature (https://flywaydb.org/documentation/scriptconfigfiles).

So if I don't have any config files for my scripts `getAbsolutePathOnDisk` is always called which seems to be a real bottleneck under Windows. 

Is it possible to deactivate the "Script Config Files"- Feature for Flyway?

Lyeeedar:
In Flyway 6.5.5 we pushed a fix that should solve the slowdown, could you try that and if it doesnt work make a new issue?

Issue #2908, opened 2020-08-24
Cleanup unused imports, formatting and codesmells


Jolley71717:
###### Which version and edition of Flyway are you using?
current master version

Reading through some of the code and there was tons of blank space, unused imports and a few code-smells.

I use flyway and would like to leave it a little cleaner than I found it.
Following the guidelines for submitting new code and adding an issue was the first step

Lyeeedar:
The blank lines and unused imports are for code found only in the pro and enterprise editions of flyway. Sadly there is no way to easily remove these for the community edition, so the code will have some oddities as a result.

Thank you for your interest though! Guidelines for how to contribute can be found here: https://flywaydb.org/documentation/contribute/

PR #2909, opened 2020-08-24
Cleaning up flyway

Jolley71717:


Lyeeedar:
As per my comment here https://github.com/flyway/flyway/issues/2908#issuecomment-679920924 we wont be accepting this PR.

Even if this was a valid change for the non-community visible sources, a massive auto-format like this would likely not be accepted. Smaller PR's formatting specific pieces of code, with a sensible justification as to why they should happen, would be much better.

PR #2934, opened 2020-09-17
Fixed broken support for CockroachDB versions >= 20.* caused by broken table already exists detection

alex-berger:
CockroachDB has a unique approach to the conception of schemas and databases which is well documented at [1]. Unfortunately, the existing implementation of [`CockroachDBTable.doExistsOnce()`](https://github.com/flyway/flyway/blob/flyway-6.5.6/flyway-core/src/main/java/org/flywaydb/core/internal/database/cockroachdb/CockroachDBTable.java#L82) is broken and will always return `false` for tables which are not part of the connection's default database (e.g. `defaultdb`).

With this PR this bug is fixed (at least for CockroachDB >= 2) by explicitly prefixing the `information_schema` schema with the table's database (schema).

[1] https://www.cockroachlabs.com/docs/stable/sql-name-resolution.htm

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/database/cockroachdb/CockroachDBDatabase.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/database/cockroachdb/CockroachDBDatabase.java
@@ -71,7 +71,7 @@ public class CockroachDBDatabase extends Database<CockroachDBConnection> {
 
     @Override
     public String getRawCreateScript(Table table, boolean baseline) {
-        return "CREATE TABLE " + table + " (\n" +
+        return "CREATE TABLE IF NOT EXISTS " + table + " (\n" +
                 "    \"installed_rank\" INT NOT NULL PRIMARY KEY,\n" +
                 "    \"version\" VARCHAR(50),\n" +
                 "    \"description\" VARCHAR(200) NOT NULL,\n" +
@@ -84,7 +84,7 @@ public class CockroachDBDatabase extends Database<CockroachDBConnection> {
                 "    \"success\" BOOLEAN NOT NULL\n" +
                 ");\n" +
                 (baseline ? getBaselineStatement(table) + ";\n" : "") +
-                "CREATE INDEX \"" + table.getName() + "_s_idx\" ON " + table + " (\"success\");";
+                "CREATE INDEX IF NOT EXISTS \"" + table.getName() + "_s_idx\" ON " + table + " (\"success\");";
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


juliahayward:
The check should only be done for the schema history table, which ought to be in the connection's default database. Do you have an example configuration which is broken by this bug that we can test against?

alex-berger:
> The check should only be done for the schema history table, which ought to be in the connection's default database. Do you have an example configuration which is broken by this bug that we can test against?

Yes, I can reproduce the problem.

Let's assume I have a CockroachDB cluster with two (datbases/schemas), once called "defaultdb" and the other "dummy" and I want to create a table in the "dummy" schema.
Then the following configurations will work (as the history table will always be created in the defaultdb schema):

`flyway migrate -url=jdbc:postgresql://192.168.10.77:26257/defaultdb -user=postgres -password=postgres`
`flyway migrate -url=jdbc:postgresql://192.168.10.77:26257/defaultdb -user=postgres -password=postgres -schemas=defaultdb,dummy`
`flyway migrate -url=jdbc:postgresql://192.168.10.77:26257/defaultdb -user=postgres -password=postgres -defaultSchema=defaultdb -schemas=defaultdb,dummy`

However, without my fix the following configuration fail, as it will try to create the history table in schema "dummy" while the `exists` check will erroneously search for the table in the "defaultdb" schema:

`flyway migrate -url=jdbc:postgresql://192.168.10.77:26257/defaultdb -user=postgres -password=postgres -schemas=dummy`

```
Flyway Community Edition 6.5.6 by Redgate
Database: jdbc:postgresql://192.168.10.77:26257/defaultdb (PostgreSQL 9.5)
Creating schema "dummy" ...
Creating Schema History table "dummy"."flyway_schema_history" ...
Creating Schema History table "dummy"."flyway_schema_history" ...
Creating Schema History table "dummy"."flyway_schema_history" ...
ERROR: 
Migration  failed
-----------------
SQL State  : 42P07
Error Code : 0
Message    : ERROR: relation "flyway_schema_history" already exists
Location   :  ()
Line       : 1
Statement  : CREATE TABLE "dummy"."flyway_schema_history" (
    "installed_rank" INT NOT NULL PRIMARY KEY,
    "version" VARCHAR(50),
    "description" VARCHAR(200) NOT NULL,
    "type" VARCHAR(20) NOT NULL,
    "script" VARCHAR(1000) NOT NULL,
    "checksum" INTEGER,
    "installed_by" VARCHAR(100) NOT NULL,
    "installed_on" TIMESTAMP NOT NULL DEFAULT now(),
    "execution_time" INTEGER NOT NULL,
    "success" BOOLEAN NOT NULL
)

Caused by: org.postgresql.util.PSQLException: ERROR: relation "flyway_schema_history" already exists
```

alex-berger:
CockroachDB version 20.2 (to be released soon) will add support for user defined schemas (beyond the default `public` schema. Therefore I revoke this PR as with that release CockroachDB should behave the same as PostgreSQL.

PR #2936, opened 2020-09-17
Updating postgres version because of XXE vulnerability

sstrang:
Versions of the postgres jdbc driver prior to 42.2.13 have an XXE vulnerability, according to this: https://nvd.nist.gov/vuln/detail/CVE-2020-13692

Would it be possible to bump the version to 42.2.13 to address this?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -175,7 +175,7 @@
         <version.mysql>8.0.17</version.mysql>
         <version.mssql-jdbc>7.2.0.jre8</version.mssql-jdbc>
         <version.oracle>19.6.0.0</version.oracle>
-        <version.postgresql>42.2.12.jre6</version.postgresql>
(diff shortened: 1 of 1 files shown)


bradparks:
It would be great to have this done, as it's a rated as a high risk, and this version bump was simply to address this issue.
https://github.com/pgjdbc/pgjdbc/blob/master/CHANGELOG.md#42213-2020-06-04

In their own words, 

> The primary reason to release this version and to continue the 42.2.x branch is for CVE-2020-13692. Reported by David Dworken, this is an XXE and more information can be found here. Sehrope Sarkuni reworked the XML parsing to provide a solution in commit 14b62aca4.

juliahayward:
Thankyou very much! Merging in to the v6 stream. Flyway v7 (due for release at the end of the month) will use the 42.2.14 version of the driver.

PR #2920, opened 2020-09-05
Configurable s3 client

rgbrizzlehizzle:
Added ability to provide a custom configured S3 Client, and also removed the block explicitly requiring AWS creds provided via env, per https://docs.aws.amazon.com/sdk-for-java/v2/developer-guide/credentials.html I don't believe that is strictly necessary.  Better to leave it up to the developer on how they handle authentication to S3.

PR #2946, opened 2020-09-23
Configurable s3 client

rgbrizzlehizzle:


rgbrizzlehizzle:
reopening https://github.com/flyway/flyway/pull/2920 off of master

rgbrizzlehizzle:
not sure what the flaky issue is here but if you look at #2920 this is the same code, just failed CI for some reason

juliahayward:
Hi! Many thanks for the contribution. As it stands, I think it creates an awkward dependency between Configuration (which is intended just to be a place to read environment and config files) and the notion of an S3Client. I assume you're using the Flyway Java API rather than the command line - could you give us a quick sample of how you would invoke Flyway with a different S3Client?

rgbrizzlehizzle:
Yes agreed, overall I'm not even too happy with the added params caused by this implementation I've added.  The code kinda smells, I'm all for other ways that would achieve this functionality.  

However, I *do* think it is crucial to allow the S3 client to be configurable at runtime.  For example, if a user is using [Localstack](https://github.com/localstack/localstack), then you need to use a modified S3 client.  

Also, in general, AWS offers a lot of flexibility on client setup/instatiation, and to have flyway kinda just ignore that and force developers down one path (use generic client or bust) seems very limiting IMO.  

My thinking here was that at runtime, the user could inject their app specific S3 client to Flyway, ideally making flyway a bit lighter on the S3 coupling side, as it is not really Flyways "job" per-say to know the specifics of how an S3 connection is being made, simply that a client has been provided.  

So (and forgive me, I haven't actually looked at this code in probably a month) but something along the lines of 

```
new Flyway(new FlywayConfig()
  .s3client(myCustomClient)
  ...
  .build()
)
```

seems like a pretty run of the mill use case that users would want to be able to leverage

DoodleBobBuffPants:
Thanks for the PR, however we have ourselves implemented the ability to pass in a custom client via the API

rgbrizzlehizzle:
Works for me, cheers!

juliahayward:
Just as an aside, after sleeping on it I've decided to keep the fluent API too - my first draft just looked horribly inconsistent!

PR #2956, opened 2020-09-30
outputQueryResults into the fluent configuration for Ant plugin

ttulka:
To offer the outputQueryResults configuration option, I need the access to the setter.
For some unclear reason the setter was private.

Please consider making it public as done in this pull request. Otherwise, I can't put it into the Ant plugin which would make it impossible even for the Teams user to use the option from the plugin.

Thanks!


MikielAgutu:
Thanks for the contribution. We'll make this public, although we can implement it ourselves.

ttulka:
ok, not only public, a new method must be put into the fluent API as well (see my commit).

PR #2964, opened 2020-10-04
Fix #2957: make software.amazon.awssdk:s3 optional in POM

### LINKED ISSUE #2957: Can Flyway 7.0.0 be used without AWS SDK and related dependencies?
Flyway 7.0.0  includes the AWS S3 SDK as a dependency. The AWS S3 SDK includes a large number of other dependencies making the footprint of Flyway 7.0.0 significantly larger than the previous release.

Having a dependency on AWS S3 SDK seems uncessary if you are not using AWS. Can Flyway 7.0.0 be used without this dependency?

vorburger:
see #2957

@MikielAgutu

### DIFF
--- a/flyway-core/pom.xml
+++ b/flyway-core/pom.xml
@@ -51,6 +51,7 @@
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>s3</artifactId>
+            <optional>true</optional>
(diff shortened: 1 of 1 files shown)


DoodleBobBuffPants:
Thanks for the PR, however we have implemented this ourselves

Issue #2966, opened 2020-10-06
Password masking when using FLYWAY_URL


dcarley:
###### Which version and edition of Flyway are you using?

```
$ docker run --rm -ti flyway/flyway:7 -v
Flyway Community Edition 7.0.0 by Redgate
```

###### If this is not the latest version, can you reproduce the issue with the latest one as well?

N/A

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin)

CLI

###### Which database are you using (type & version)?

```
$ docker run --rm -ti postgres:9.6 -V
postgres (PostgreSQL) 9.6.19
$ docker run --rm -ti -e POSTGRES_USER=test_user -e POSTGRES_PASSWORD=test_pass -e POSTGRES_DB=test_db -p 5432:5432 --network flyway --name postgres postgres:9.6
```

###### Which operating system are you using?

Linux (via Docker on Mac, in this reproduction case)

###### What did you do?

1. Ran a migration with a password in `FLYWAY_URL` and debug mode enabled:
    ```
    $ docker run --network flyway --rm -ti -v "$(pwd)/fixtures:/flyway/sql" -e FLYWAY_URL='jdbc:postgresql://postgres:5432/test_db?user=test_user&password=test_pass' flyway/flyway:7 migrate -X
    ```
1. Ran a migration with an incorrect password in `FLYWAY_URL`:
    ```
    $ docker run --network flyway --rm -ti -v "$(pwd)/fixtures:/flyway/sql" -e FLYWAY_URL='jdbc:postgresql://postgres:5432/test_db?user=test_user&password=wrong_pass' flyway/flyway:7 migrate
    ```

###### What did you expect to see?

The password to be masked out in the debug output and connection errors, in the same way as it is when using `FLYWAY_PASSWORD`:
```
DEBUG: flyway.password -> *********
```

###### What did you see instead?

1. Debug mode:
    ```
    DEBUG: Multiple databases found that handle url 'jdbc:postgresql://postgres:5432/test_db?user=test_user&password=test_pass'. CockroachDB, PostgreSQL
    DEBUG: Multiple databases found that handle url 'jdbc:postgresql://postgres:5432/test_db?user=test_user&password=test_pass'. CockroachDB, PostgreSQL
    DEBUG: Using configuration:
    DEBUG: flyway.jarDirs -> /flyway/jars
    DEBUG: flyway.locations -> filesystem:sql
[...]

PR #2973, opened 2020-10-09
[issue/2966]: Mask password in jdbc url

OskarsPakers:
Implements #2966 by masking password while logging jdbc URL


DoodleBobBuffPants:
Thanks for the PR, however we are already implementing this ourselves for a future release, so will close this

PR #2976, opened 2020-10-12
Add Support for CockroachDB 20.2

alex-berger:
This PR adds support for the upcoming CockroachDB version 20.2, which will support custom schemas (beside the default `public` schema).

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/database/cockroachdb/CockroachDBSchema.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/database/cockroachdb/CockroachDBSchema.java
@@ -38,6 +38,11 @@ public class CockroachDBSchema extends Schema<CockroachDBDatabase, CockroachDBTa
     final boolean cockroachDB1;
 
     /**
+     * Is this CockroachDB >= 20.2
+     */
+    final boolean hasSchemaSupport;
+
+    /**
      * Creates a new CockroachDB schema.
      *
      * @param jdbcTemplate The Jdbc Template for communicating with the DB.
@@ -47,6 +52,7 @@ public class CockroachDBSchema extends Schema<CockroachDBDatabase, CockroachDBTa
     CockroachDBSchema(JdbcTemplate jdbcTemplate, CockroachDBDatabase database, String name) {
         super(jdbcTemplate, database, name);
         cockroachDB1 = !database.getVersion().isAtLeast("2");
+        hasSchemaSupport = database.getVersion().isAtLeast("20.2");
     }
 
     @Override
@@ -60,6 +66,9 @@ public class CockroachDBSchema extends Schema<CockroachDBDatabase, CockroachDBTa
     }
 
     private boolean doExistsOnce() throws SQLException {
+        if ( hasSchemaSupport ) {
+            return jdbcTemplate.queryForBoolean("SELECT EXISTS ( SELECT 1 FROM information_schema.schemata WHERE schema_name=? )", name);
+        }
         return jdbcTemplate.queryForBoolean("SELECT EXISTS ( SELECT 1 FROM pg_database WHERE datname=? )", name);
     }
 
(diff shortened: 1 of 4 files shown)


flyway-core/src/main/java/org/flywaydb/core/internal/database/cockroachdb/CockroachDBSchema.java
```diff
@@ -81,19 +90,32 @@ private boolean doEmptyOnce() throws SQLException {
                     "  WHERE table_schema=?" +
                     "  AND table_type='BASE TABLE'" +
                     ")", name);
+        } else if ( ! hasSchemaSupport ) {
+            return !jdbcTemplate.queryForBoolean("SELECT EXISTS (" +
+                    "  SELECT 1" +
+                    "  FROM information_schema.tables " +
+                    "  WHERE table_catalog=?" +
+                    "  AND table_schema='public'" +
```
rafiss, line 98: since the two queries are mostly the same, I wonder if it would be better to build up the query with string concatenation. like:

```
String query = "SELECT EXISTS (" +
  "  SELECT 1" +
  "  FROM information_schema.tables " +
  "  WHERE table_catalog=?";
if (hasSchemaSupport) {
  query += " WHERE table_schema=?";
}
query += "  AND table_type='BASE TABLE'" +
// and so on.
```

flyway-core/src/main/java/org/flywaydb/core/internal/database/cockroachdb/CockroachDBTable.java
```diff
@@ -70,39 +71,60 @@ public Boolean call() throws SQLException {
...
+        } else if ( !schema.hasSchemaSupport ) {
+            return jdbcTemplate.queryForBoolean("SELECT EXISTS (\n" +
+                    "   SELECT 1\n" +
+                    "   FROM   information_schema.tables \n" +
+                    "   WHERE  table_catalog = ?\n" +
+                    "   AND    table_schema = 'public'\n" +
+                    "   AND    table_name = ?\n" +
+                    ")", schema.getName(), name);
+        } else {
+            // There is a bug in CockroachDB v20.2.0-beta.* which causes the string equality operator to not work as
```
rafiss, line 90: This will be fixed in the official 20.2.0 release. So I guess it's a judgment call if this code needs to work with the beta release or not.

alex-berger: To properly test the CockroachDB beta release we have to make sure all tooling around it (like for example Flyway or JOOQ) work as expected, therefore it is important that this fix also works with the (latest) beta release. Furthermore, my company is currently relying on features only available in CockroachDB 20.2 (e.g. custom schemas and partial indexes), so we need this to work with the beta release.

flyway-core/src/main/java/org/flywaydb/core/internal/database/cockroachdb/CockroachDBTable.java
```diff
@@ -70,39 +71,60 @@ public Boolean call() throws SQLException {
...
-                "   AND    table_name = ?\n" +
-                ")", schema.getName(), name);
     }
 
     @Override
     protected void doLock() throws SQLException {
-        if (lockDepth == 0) {
-            insertRowLock.doLock(jdbcTemplate, database.getInsertStatement(this), database.getBooleanTrue());
+        if ( database.getVersion().isAtLeast("20.1") ) {
+            jdbcTemplate.execute("SELECT * FROM " + this + " FOR UPDATE");
```
rafiss, line 107: I am not sure if this will work as desired, since the migration is not run inside of a transaction. (the lock from a SFU update is only held for the duration of a transaction.)

flyway-core/src/main/java/org/flywaydb/core/internal/database/cockroachdb/CockroachDBTable.java
```diff
@@ -30,6 +30,7 @@
  * carrying out a migration.
```
rafiss, line 30: note: if the locking strategy does change, the comment here should be updated. But I'm not sure if the locking can be changed right now.

flyway-core/src/main/java/org/flywaydb/core/internal/database/cockroachdb/CockroachDBTable.java
alex-berger: OK, I will revert that optimization and always use the `insertRowLock` instead.

flyway-core/src/main/java/org/flywaydb/core/internal/database/cockroachdb/CockroachDBSchema.java
alex-berger: I had the same thoughts, but then I decided to prefer easy readable code with some redundancy over more complex code without redundancy.

vy-ton:
@juliahayward Can you help prioritize/review this?

juliahayward:
Certainly, we'll need to expand our test suite to pull in a 20.2 Docker container and make sure everything is still OK. 

One question arising from that - in `doDropOnce()`, it appears `DROP SCHEMA IF EXISTS ...` needs the `CASCADE` option set. Is that what you'd expect or is there something unusual in our test?

alex-berger:
@juliahayward yes, your are right. I will add `CASCADE`. Thanks for pointing out.

DoodleBobBuffPants:
Thanks for the PR! This will be in Flyway 7.1

While we did redo some of it, nonetheless, welcome to the hall of fame!

alex-berger:
@DoodleBobBuffPants Thank you very much for integrating this changes into Flyway. I just tested 7.1.0 with our CockroachDB Flyway migrations and it works as expected.

PR #3014, opened 2020-11-20
Move the Resource class to the public API package

### LINKED ISSUE #3013: Move the Resource class into the API package
Sort of as a follow on to #2822, I'd like to suggest that the ResourceProvider class that was recently moved into the public API package be refactored to return a kind of Resource object that also exists in the public API package. It currently returns LoadableResource, which is in the org.flywaydb.core.internal package.

Just to add some context, my organization has some custom Flyway code that has used classes out of this internal package for quite some time, and we're making an effort to refactor our code to only use classes in the API package so that fewer things will break when we upgrade Flyway.

The LoadableResource class itself doesn't look particularly useful, since all classes that implement the Resource interface also extend from this threadbare abstract class. It shouldn't be too difficult or disruptive to absorb the LoadableResource class into the Resource interface, and move that interface to the public API package.

That would require a non-backwards compatible change to the public facing ResourceProvider class, but it would theoretically prevent more breaking changes in the future.

I have a PR that I'm happy to submit if it would be welcome.

Thanks!

ericparton:
Fixes #3013


DoodleBobBuffPants:
Thanks for the PR, however if we decided to make such a change to the API we would do it ourselves, so closing this

Issue #3084, opened 2021-02-16
Gradle plugin without java plugin ignores configurations classpath


TarodBOFH:
###### Which version and edition of Flyway are you using?

Community, 7.5.3 :
```
plugins {
    id("org.flywaydb.flyway") version "7.5.3"
}
``` 

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin)

Gradle plugin

###### Which database are you using? (Type & version)

It doesn't matter, but the example in the attached git repo is with MySQL 8.0.23

###### Which operating system are you using?

It doesn't matter. It is a Gradle-plugin bug... anyway, it happens on official Gradle docker images, windows, MacOS... 

###### What did you do? (Please include the content causing the issue, any relevant configuration settings, the SQL statement(s) that failed (if any), and the command you ran)

I added just flyway to an empty project and added mysql driver to a configurations block. The driver is not loaded during flyway tasks inside gradle-plugin

###### What did you expect to see?
A connection timeout instead of a missing driver.
```
Executing task 'flywayInfo'...

> Task :flywayInfo FAILED
1 actionable task: 1 executed

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':flywayInfo'.
> Error occurred while executing flywayInfo
  
  Unable to obtain connection from database (jdbc:mysql://localhost?autoReconnect=true&useSSL=false&useUnicode=yes&characterEncoding=UTF-8) for user 'root': Could not create connection to database server. Attempted reconnect 3 times. Giving up.
...
```

###### What did you see instead?
A missing driver error:

```
Executing task 'flywayInfo'...

> Task :flywayInfo FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':flywayInfo'.
> Error occurred while executing flywayInfo
  Unable to instantiate JDBC driver: com.mysql.cj.jdbc.Driver => Check whether the jar file is present
  Unable to instantiate class com.mysql.cj.jdbc.Driver : com.mysql.cj.jdbc.Driver
  com.mysql.cj.jdbc.Driver
```

I've created a minimal gradle project that reproduces both behaviors in two git branches at https://github.com/TarodBOFH/flyway-without-java-bug

I'll be submitting a pull request with the fix as soon as I finish writing the issue.

DoodleBobBuffPants:
Closing this to continue on the PR instead

PR #3085, opened 2021-02-16
Fix  https://github.com/flyway/flyway/issues/3084

TarodBOFH:
Allow to load configuration artifacts even when it is not a java project (i.e. one might be running flyway as a submodule without any kind of java dependencies)

I'll upload tomorrow at https://github.com/TarodBOFH/flyway-without-java-bug a version that takes the plugin snapshot from maven local instead with the path applied.

### DIFF
--- a/flyway-gradle-plugin/src/main/java/org/flywaydb/gradle/task/AbstractFlywayTask.java
+++ b/flyway-gradle-plugin/src/main/java/org/flywaydb/gradle/task/AbstractFlywayTask.java
@@ -683,10 +683,14 @@ public abstract class AbstractFlywayTask extends DefaultTask {
         if (extension.configurations != null) {
             return extension.configurations;
         }
-        if (getProject().getGradle().getGradleVersion().startsWith("3")) {
-            return DEFAULT_CONFIGURATIONS_GRADLE3;
+        if (isJavaProject()) {
+            if (getProject().getGradle().getGradleVersion().startsWith("3")) {
+                return DEFAULT_CONFIGURATIONS_GRADLE3;
+            }
+            return DEFAULT_CONFIGURATIONS_GRADLE45;
+        } else {
+            return new String[0];
         }
-        return DEFAULT_CONFIGURATIONS_GRADLE45;
     }
 
     /**
--- a/flyway-gradle-plugin/src/main/java/org/flywaydb/gradle/task/AbstractFlywayTask.java
+++ b/flyway-gradle-plugin/src/main/java/org/flywaydb/gradle/task/AbstractFlywayTask.java
@@ -606,8 +606,9 @@ public abstract class AbstractFlywayTask extends DefaultTask {
             Set<URL> extraURLs = new HashSet<>();
             if (isJavaProject()) {
                 addClassesAndResourcesDirs(extraURLs);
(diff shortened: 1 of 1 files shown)


TarodBOFH:
The branch https://github.com/TarodBOFH/flyway-without-java-bug/tree/fix-applied has the fix uploaded as a local mvn repo and is able to load the drivers despite no java plugin loaded.

DoodleBobBuffPants:
Could you explain the reason behind the second commit and why it is needed/desired in this PR? It doesn't seem like it is necessary in order to solve the original issue - if it isn't, would you be able to remove the last commit and submit it as a separate PR explaining its purpose?

TarodBOFH:
> Could you explain the reason behind the second commit and why it is needed/desired in this PR? It doesn't seem like it is necessary in order to solve the original issue - if it isn't, would you be able to remove the last commit and submit it as a separate PR explaining its purpose?

Sure, really sorry. I went to bed and realized a mistake:

```
    private void addConfigurationArtifacts(String[] configurations, Set<URL> urls) throws IOException {
        for (String configuration : configurations) {
            getLogger().debug("Adding configuration to classpath: " + configuration);
            ResolvedConfiguration resolvedConfiguration =
                    getProject().getConfigurations().getByName(configuration).getResolvedConfiguration();
            for (ResolvedArtifact artifact : resolvedConfiguration.getResolvedArtifacts()) {
                URL artifactUrl = artifact.getFile().toURI().toURL();
                getLogger().debug("Adding artifact to classpath: " + artifactUrl);
                urls.add(artifactUrl);
            }
        }
    }
```

That is the method that resolves the artifacts of given configurations. Since on the first commit I moved out of the java plugin. 

I realized that with no custom configurations (env, properties or gradle ones), then the default configurations are there (compile, runtime, etc.). In that corner case, the build will fail because `                    getProject().getConfigurations().getByName(configuration).getResolvedConfiguration();` cannot find those configurations that are added by the java plugin and throws an exception. This second commit fixes that corner case.

TarodBOFH:
I would like to propose an iteration over this change. As per gradle recomendations, convention over configuration, flyway-gradle-plugin could use two configurations:
- `flywayImplementation` -> to declare libs and drivers to be loaded
- `flywayRuntime` (extends `flywayImplementation`), which can be resolved 

That way a typical configuration would be 
```
dependencies {
   flywayImplementation(".... your drivers and libs here")
}

---- flyway-gradle-plugin pseudo-code

val flywayImplementation = configurations.creating("flywayImplementation") { isCanBeResolved = false }
configurations.create("flywayRuntine") { 
    extends(flywayImplementation) 
    isCanBeConsumed = false
}
```

That way, only flywayRuntime is added to flyway classpath instead of the full java plugin classpath.

The way JavaPlugin does this is pretty straighforward (has a bucket for deps, then it is added to runtime) but it would render incompatible changes with the current version.

DoodleBobBuffPants:
We'll keep that proposition in mind but won't be able to commit to it, however the changes in this PR seem sensible in extending the Gradle plugin's compatibility

DoodleBobBuffPants:
This has been mirrored and merged, so closing

And welcome to the hall of fame 🎉

PR #3125, opened 2021-03-26
Drop jre7 classifier from postgresql JDBC driver version

pzygielo:
As mentioned in https://github.com/flyway/flyway/commit/a912d87eb5dacb2a4b83bbed5f42e03b71519419#r48670327 - I'd rather have this driver version...


DoodleBobBuffPants:
Thanks for the PR, however we are already in the process of considering this change, so closing

PR #3127, opened 2021-03-26
Checksum improvements

Kosta-Git:
# Checksum case insensitive & spacing insensitive

## Summary

The checksum was a basic `CRC32` checksum which means that **any** spacing **changes** or upper/lower casing SQL keywords would **invalidate the checksum**.

This pull request contains some code that allows the user to change the formatting of the file ( newlines, spacing, casing ) and keeping the same checksum.

The checksum code proposal tokenizes the SQL file and outputs a string containing normalized SQL ( every SQL keyword is uppercased and no unnecessary spacing ), this can be thought of processing the file to a sematic SQL string which would result in the checksum of the SQL Logic instead of the actual file.

All the SQL keywords are configurable inside the `SqlKeywords` class which allows to add future keywords easily.

## Examples

Here is an example showing what the changes would allow user to do, in this case I wrote a small example, then configured my prettifier and ran it. Same SQL logic, the checksum end up being the same for both files.

```sql
name: V2020_1_example.sql
checksum: 538194993

create table if not exists EXAMPLE
(
    example_id bigserial not null constraint example_pk primary key,
    example_name varchar not null
);
```

```sql
name: V2020_1_example.sql
checksum: 538194993

CREATE TABLE IF NOT EXISTS EXAMPLE
(
    example_id bigserial not null 
                       constraint 
                       example_pk
                       primary key,

    example_name VARCHAR NOT NULL
);
```

--- 

If we do an insert statement the fields as well as the values should be case sensitive, and for the string literal it should also be space sensitive.

```sql
name: V2020_2_example_two.sql
checksum: -179199000

insert into public.EXAMPLE (example_id, example_name) values (1, 'test') on conflict do nothing;
```

Now if we change the content of the string

```sql
name: V2020_2_example_two.sql
checksum: 1212101293

insert into public.EXAMPLE (example_id, example_name) values (1, 'TEST') on conflict do nothing;
```

We can see the checksum are not matching, but if we keep the same content and just change the formating

```sql
name: V2020_2_example_two.sql
checksum: -179199000

INSERT INTO 
public.EXAMPLE 
(example_id, example_name) 
VALUES (1, 'test') 
ON CONFLICT DO NOTHING;
```

We get the same checksum as the first try.


DoodleBobBuffPants:
While we appreciate the PR, we cannot accept this at the moment

However, we will consider this idea for the next major version as it is a big breaking change

PR #2981, opened 2020-10-20
Clickhouse replicated

bennyhils:

PR #3134, opened 2021-04-06
Clickhouse support

strug:
Use pr https://github.com/flyway/flyway/pull/2981 - update for new scheme implementation. Fix some problem with distributed logic, when more one instance application.


GordeevArt:
In flyway replicated table creation zookeeper path should be "/clickhouse/{installation}/{cluster}/tables/{shard}/{database}/{table}" instead of "/clickhouse/tables/{shard}/" + tableName. It prevents conficts with mutliple clickhouse installations with flyway in one zookeeper.

den-crane:
> In flyway replicated table creation zookeeper path should be

should be???
99% of CH users don't have macros {cluster}
100% of CH users don't have macros {installation}
It simply won't work.

GordeevArt:
> > In flyway replicated table creation zookeeper path should be
> 
> should be??? 99% of CH users don't have macros {cluster} 100% of CH users don't have macros {installation} It simply won't work.

Maybe add **zookeeper.path** parameter?

den-crane:
BTW it's a Russian С not С in the PR name "Сlickhouse support"
That's confusing , I copy paste Сlickhouse word and was unable to find it.

johnny:
What is the status of this PR and the chances of it getting merged? We'd like to use flyway for clickhouse.

harshvikramsingh:
Hi, we are also waiting for this change. Can this be please merged if no more issues?

flyway-commandline/pom.xml
```diff
@@ -185,6 +185,11 @@
             <artifactId>lombok</artifactId>
         </dependency>
 
+        <dependency>
```
TheHett, line 188: this driver is obsolete

Barry-RG:
Clickhouse has been added via subsequent PRs

PR #3163, opened 2021-04-29
Inline OracleConnection method usage

gastaldi:
GraalVM does not like when a class that is not in the classpath is declared in the method signature.


ilopmar:
I opened #3175 because and error related to this trying to upgrade Micronaut Flyway.
I've tried this patch locally and it works perfectly :100: 

Do you have any ETA to get this merged and release Flyway 7.9.1 with it? Thanks!

DoodleBobBuffPants:
Thanks for the PR, and welcome to the hall of fame!

We've duplicated and merged this PR, so closing now, and it will be available in the next release

ilopmar:
That was fast, thanks. Waiting for next release now :+1:

PR #3166, opened 2021-05-06
Use Timestamp with Time zone in PostgreSQL

gastaldi:
See https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_timestamp_.28without_time_zone.29


DoodleBobBuffPants:
Closing unless we can link this to an active issue as it will be a breaking change to do this

PR #3185, opened 2021-05-29
Allow creation of schema history table on non-empty schemas

### LINKED ISSUE #3184: Allow creation of schema history table on non-empty schemas

7.9.1

Java API

Postrgres 11

MacOS

We have libraries that need to do a separate migration flow in the same database as the main application. The obvious solution is to have libraries run their own Flyway flow with an alternate history table name. However, this is currently not possible with Flyway because Flyway will throw an exception when the schema is not empty and the history table is missing.

PR: https://github.com/flyway/flyway/pull/3185

Randgalt:
Allow history table creation on non-empty schemas via a new
configuration. This allows library writers to use Flyway in
an neutral/independent manner without disrupting the main
applications Flyway usage.

A library can set a different history table name, set the
new configuration and run a separate, internal Flyway
instance to do the library's separate migrations.

This is a very safe change as it's opt-in. The default behavior 
remains the same. Users must enable this new behavior.

Note: I don't see any tests in the project but I did test locally 

Closes #3184

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/Flyway.java
@@ -180,9 +180,11 @@ public class Flyway {
 
                 if (!schemaHistory.exists()) {
                     List<Schema> nonEmptySchemas = new ArrayList<>();
-                    for (Schema schema : schemas) {
-                        if (schema.exists() && !schema.empty()) {
-                            nonEmptySchemas.add(schema);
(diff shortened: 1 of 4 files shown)


DoodleBobBuffPants:
Closing this PR for now

As per the ongoing discussion on https://github.com/flyway/flyway/issues/3184, `baselineOnMigrate` should be used in this case, and we'll aim to resolve that issue and return to this PR if needed

PR #3205, opened 2021-06-11
Support H2 2.0.201 (will break existing H2 1.4.200 support)

craigraw:
These changes allow flyway to work with the latest HEAD from H2. 

Merging these changes will break H2 1.4.200 support! However they may be useful as a starting point, or for others who are looking to migrate to the newer H2 releases. If there is a way to do this supporting both versions, please let me know.


juliahayward:
I will look at an implementation that keeps compatibility for both. 

Re. the changes to H2Schema, are these code fragments removed because the corresponding database features have been removed?

craigraw:
> I will look at an implementation that keeps compatibility for both.

Thanks!

> Re. the changes to H2Schema, are these code fragments removed because the corresponding database features have been removed?

As far as I can see, they have either been removed, or the implementation has changed in a way that no longer requires them. There are no similar substitutes (again, as far as I can see).

juliahayward:
Many thanks. Closing this PR, the next release will have both v1.x and v2 support with auto-detection.

Issue #3211, opened 2021-06-18
Add string concatenation around all LOG.debug() calls with string concatenation


keesvandieren:
###### Which version and edition of Flyway are you using?

7.3.0

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin)

Java API

###### Which database are you using? (Type & version)

Not relevant

###### Which operating system are you using?

N/A

###### What did you expect to see?

I see that many strings concatenations are done in LOG.debug calls without isDebugEnabled() condiation around it

###### What did you see instead?

isDebugEnabled() statement around all LOG.debug statements that do string concatenation

DoodleBobBuffPants:
Closing as per https://github.com/flyway/flyway/pull/3212

PR #3212, opened 2021-06-18
No concatenation for LOG.debug() calls if DEBUG disabled

keesvandieren:
Fix for #3211.


DoodleBobBuffPants:
Thanks for the PR, however we won't be merging it as it bloats the code for little perceived value - unless you are able to prove there is a significant performance change?

PR #3244, opened 2021-07-30
Don't use user-provided ClassLoader for DatabaseTypeRegister

jamiecounsell:
As seen in flyway#3177, DatabaseTypeRegister shouldn't really be using the classloader provided in the configuration. In OSGI, this causes Flyway to be unable to find the database drivers, since they are only visible to the Flyway bundle and the user's bundle is only really needed to provide resources like SQL migrations.

I can't really see a good use case for using a custom classLoader in the DatabaseTypeRegister but let me know if I misunderstand that. If we need this, then we should probably add a configuration option that allows people to register their own database types if they want to bring their own.


DoodleBobBuffPants:
Thanks for the PR! We have taken it and implemented a more complete version, so closing this

PR #3249, opened 2021-08-10
Supports OpenTelemetry JDBC prefix.

donbeave:
This PR adds support for `jdbc:otel:` prefix, which can be used by OpenTelemetry Manual Instrumentation: https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/jdbc/library.


donbeave:
@DoodleBobBuffPants Sorry for tagging you personally here, can you help me to get some information, is these changes is ok to be merged or these changes are considered not applicable for the flyway and should be resolved in the application directly? For me, it looks like the same thing as supporting P6Spy, and because support for p6spy prefixes is already integrated into the flyway library, I think support for OTEL (OpenTelemetry) prefixes must be here as well.

DoodleBobBuffPants:
We are unable to prioritize this since there's not enough interest as far as we can tell, and we'd also want to do some rearchitecting so that adding support for things like P6Spy and OTEL doesn't require modifying every database type class

donbeave:
Maybe in this case it's better for me to wait until a specific version and come back here again?

PR #3253, opened 2021-08-18
Correctly specify rethrow on error during lock template execution

FylmTM:
We have encountered an error "Unable to release PostgreSQL advisory lock" during our application start.

Looking through logs it seems that everything worked fine until some point.
We have suspicion that Flyway "swallows" original problem that is "SQLException", by throwing new exception in `finally` block.

This PR specifies `rethrow` (in the same way as done in next `catch` block in same method), so in `unlock` it could be correctly picked up.


DoodleBobBuffPants:
Could you sign the CLA?

FylmTM:
@DoodleBobBuffPants sorry.

Already signed CLA before. But used wrong email on commit.
Fixed now!

DoodleBobBuffPants:
Thanks! We have cloned the PR and have it running through our pipeline, so closing this

PR #3254, opened 2021-08-19
Make CockroachDBConnection constructor protected

electrum:


DoodleBobBuffPants:
Thanks for the PRs, we have them running through our pipeline, so closing these

PR #3255, opened 2021-08-19
Make CockroachDBSchema constructor public

electrum:
This makes it easier to subclass or replace `CockroachDBConnection`.


DoodleBobBuffPants:
Thanks for the PRs, we have them running through our pipeline, so closing these

Issue #3266, opened 2021-08-28
Java callbacks: ignore abstract classes


kgroble:
###### Which version and edition of Flyway are you using?

7.5.0

###### If this is not the latest version, can you reproduce the issue with the latest one as well? (Many bugs are fixed in newer releases and upgrading will often resolve the issue)

Yes, reproduced in 7.14.0

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin)

Java API

###### Which database are you using? (Type & version)

PostgreSQL 13.3

###### Which operating system are you using?

Mac, Linux

###### What did you do? (Please include the content causing the issue, any relevant configuration settings, the SQL statement(s) that failed (if any), and the command you ran)

Run migrations with an abstract class implementing `Callback` on the classpath provided to the `callbacks` configuration.

###### What did you expect to see?

Abstract class is skipped, as abstract classes are for migrations (#326).

###### What did you see instead?

Flyway attempts to instantiate the abstract class and throws an exception.

kgroble:
I have prepared a fix for this issue at: https://github.com/flyway/flyway/pull/3267

PR #3267, opened 2021-08-28
Fix for #3266: do not attempt to instantiate abstract callbacks

kgroble:
Skip abstract classes when loading callbacks. Since `ClassUtils.loadClass` catches all throwables, this also "quietly" skips classes for which instantiation was attempted but failed (a warning is logged, but such cases would previously cause `FluentConfiguration::callbacks` to error out).

I have tested this with my application, and I can now have my abstract callback class in the same package as its concrete implementations. I did not see anywhere to add automated test coverage for this, but I would be happy to if there is such a place.


DoodleBobBuffPants:
Thanks for the PR! We have cloned it and have it running through our pipeline, so closing this one

PR #3337, opened 2021-12-05
Fixed "Flyway using incorrect metadata column names for latest h2 db (2.0.202)" (#3334)

### LINKED ISSUE #3334: Flyway using incorrect metadata column names for latest h2 db (2.0.202)
8.1.0

N/A

Java API

h2 2.0.202

OS X

1. Created h2 tcp server
2. Ran flyway clean 
```
        Flyway flyway;
        private String getJdbcUrl(String tcpUrl) {
            return "jdbc:h2:" + tcpUrl + "/./database_name;MODE=ORACLE;DEFAULT_NULL_ORDERING=HIGH";
        }
        ...
        @Override
        protected void before() throws Throwable {

            h2Server = Server.createTcpServer("-baseDir", "target", "-tcpAllowOthers", "-ifNotExists");
            h2Server.start();

            flyway = Flyway.configure()
                    .dataSource(getJdbcUrl(h2Server.getURL()),"...", "...")
                    .baselineOnMigrate(true)
                    .outOfOrder(true)
                    .load();

            flyway.clean();

        }
```
 
I expected flyway to successfully clean the h2 database

I received the following exception:
```
org.flywaydb.core.internal.exception.FlywaySqlException: Unable to determine H2 compatibility mode
-----------------------------------------
SQL State  : 42001
Error Code : 42001
Message    : Syntax error in SQL statement "SELECT VALUE[*] FROM INFORMATION_SCHEMA.SETTINGS WHERE NAME = 'MODE'"; expected "DISTINCT, ALL, ,, NOT, EXISTS, INTERSECTS, UNIQUE"; SQL statement:
SELECT VALUE FROM INFORMATION_SCHEMA.SETTINGS WHERE NAME = 'MODE' [42001-202]
```

### Notes:

It appears that there is a call to determine compatibility mode when the H2Database is constructed.

https://github.com/flyway/flyway/blob/d1bd7ca8109b344774443b724f380fd1d157e80f/flyway-core/src/main/java/org/flywaydb/core/internal/database/h2/H2Database.java#L59-L63)

The method `determineCompatibilityMode` relies on the field `requiresV2MetadataColumnNames`.  This field is initialized to false. Therefore the compatibility query is using the old metadata column names.
https://github.com/flyway/flyway/blob/d1bd7ca8109b344774443b724f380fd1d157e80f/flyway-core/src/main/java/org/flywaydb/core/internal/database/h2/H2Database.java#L83-L86

Also, from what I can see the only way to properly initialize this value is to call `ensureSupported` which is called after the database has been constructed.
https://github.com/flyway/flyway/blob/d1bd7ca8109b344774443b724f380fd1d157e80f/flyway-core/src/main/java/org/flywaydb/core/FlywayExecutor.java#L172-L189

tototoshi:
This fixes #3334 

The current version is using the `requiresV2MetadataColumnNames` flag to determine which columns to use, but we were getting errors because Flyway tries to get the metadata from the database to determine whether `requiresV2MetadataColumnNames` was true or false.
In this fix, the determination of requiresV2MetadataColumnNames is done in the constructor. Also, the version information in the parent class is enough for this process, so we use that for the decision.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/database/h2/H2Database.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/database/h2/H2Database.java
@@ -59,6 +59,7 @@ public class H2Database extends Database<H2Connection> {
     public H2Database(Configuration configuration, JdbcConnectionFactory jdbcConnectionFactory, StatementInterceptor statementInterceptor) {
         super(configuration, jdbcConnectionFactory, statementInterceptor);
 
+        requiresV2MetadataColumnNames = super.determineVersion().isAtLeast("2.0.0");
         compatibilityMode = determineCompatibilityMode();
     }
 
@@ -112,7 +113,6 @@ public class H2Database extends Database<H2Connection> {
 
         recommendFlywayUpgradeIfNecessary("2.0.201");
         supportsDropSchemaCascade = getVersion().isAtLeast("1.4.200");
-        requiresV2MetadataColumnNames = getVersion().isAtLeast("2.0.0");
     }
 
     @Override


Barry-RG:
A PR to upgrade flyway to be 2.0.202 compatible is already in our pipeline so will close this PR here

Issue #3348, opened 2021-12-14
Support for SQL Server BEGIN DIALOG and BEGIN CONVERSATION TIMER


jasouyris:
###### Which version and edition of Flyway are you using?
8.2.2 Community Edition.

###### If this is not the latest version, can you reproduce the issue with the latest one as well? (Many bugs are fixed in newer releases and upgrading will often resolve the issue)
The issue is on the latest version.

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin)
Gradle plugin.

###### Which database are you using? (Type & version)
SQL Server 2012.

###### Which operating system are you using?
Windows Server 2012.

###### What did you do? (Please include the content causing the issue, any relevant configuration settings, the SQL statement(s) that failed (if any), and the command you ran)
I tried to apply a SQL Server migration that had BEGIN DIALOG and BEGIN CONVERSATION TIMER in it.

###### What did you expect to see?
I expect [BEGIN DIALOG](https://docs.microsoft.com/en-us/sql/t-sql/statements/begin-dialog-conversation-transact-sql?view=sql-server-ver15) and [BEGIN CONVERSATION TIMER](https://docs.microsoft.com/en-us/sql/t-sql/statements/begin-conversation-timer-transact-sql?view=sql-server-ver15) to adjust the block depth the same way as [BEGIN { TRAN | TRANSACTION }](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/begin-transaction-transact-sql?view=sql-server-ver15).

###### What did you see instead?
The block depth is not being adjusted so it fails.

PR #3349, opened 2021-12-14
Fix #3347 and #3348

### LINKED ISSUE #3347: NPE on TRANSACTION_REGEX.matcher(keywordText)
8.2.2 Community Edition.

The issue is on the latest version.

Gradle plugin.

SQL Server 2012.

Windows Server 2012.

A NPE is thrown in `SQLServerParser` in `TRANSACTION_REGEX.matcher(keywordText)` due to `keywordText` being null in some cases.

I'm expecting for both `isTransaction()` and `isDistributedTransaction()` to return false when `keywordText` is null.

I get a NPE.

jasouyris:


DoodleBobBuffPants:
Thanks for the PR, however we have a fix for the related issues in our pipeline so closing this

Issue #3360, opened 2021-12-20
Log4j bump to 2.17


larrydiamond:
CVE-2021-45105
More log4shell remediation
Patch forthcoming

Barry-RG:
Hi @larrydiamond 
We are trying to keep up to date with the seemingly on-going Log4Shell developments. However, as Flyway's dependency on Log4J is optional and configured by our users we will be updating as part of our usual release pipeline schedule. 
This is because the version we state is only for development proposes and is under user control when running.
Please see our [blog post](https://flywaydb.org/blog/flyway-log4j-vulnerability) for further explanation.

PR #3367, opened 2021-12-23
log4j 2.17.0

sullis:
https://logging.apache.org/log4j/2.x/security.html


DoodleBobBuffPants:
Thanks for the PR however we won't be merging this just yet

See https://github.com/flyway/flyway/issues/3360#issuecomment-998655952 for more details on why this doesn't directly Flyway

Issue #2895, opened 2020-07-30
Support for PgBouncer transaction pooling


matmannion:
###### Which version and edition of Flyway are you using?

6.5.2

###### If this is not the latest version, can you reproduce the issue with the latest one as well?
(Many bugs are fixed in newer releases and upgrading will often resolve the issue)

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin)

Java API

###### Which database are you using (type & version)?

postgres (PostgreSQL) 10.13
PgBouncer 1.14.0

###### Which operating system are you using?

CentOS Linux release 7.8.2003

###### What did you do?
(Please include the content causing the issue, any relevant configuration settings, the SQL statement that failed (if relevant) and the command you ran.)

Connected to the PostgreSQL database via PgBouncer, using transaction pooling. Flyway.migrate() is called as part of startup.

This is part of a multi-node environment where multiple nodes start simultaneously, meaning there are multiple nodes starting Flyway from the same code-base at the same time.

###### What did you expect to see?

Validation of the existing migrations and performing any missing ones.

###### What did you see instead?

Multiple nodes appear to "lock", waiting for a time-out and then dying with an error "Number of retries exceeded while attempting to acquire PostgreSQL advisory lock".

Looking at the way the code works, this is expected; Flyway grabs a _session_ lock as part of its work (https://github.com/flyway/flyway/blob/master/flyway-core/src/main/java/org/flywaydb/core/internal/database/postgresql/PostgreSQLAdvisoryLockTemplate.java#L106-L116) but PgBouncer is set up to do transaction pooling, so multiple nodes may (and do, in this case) receive the same database session for separate transactions. This then means that the session lock can't be obtained and the node fails with the above error.

In this situation (and with group=true in settings) it _may_ be more appropriate to get a transaction-level advisory lock rather than a session-level advisory lock by using `pg_try_advisory_xact_lock` instead of `pg_try_advisory_lock`, which may be sufficient to get this working with PgBouncer transaction pooling.

fml2:
> multiple nodes may (and do, in this case) receive the same database session for separate transactions. This then means that the session lock can't be obtained and the node fails with the above error.

Shouldn't it rather lead to the situation that ALL nodes successfully aquire the lock (since they work within same session as seen by the DB)? Hence, effectively, no sync would take place.

According to the docs (see [here](https://www.postgresql.org/docs/current/explicit-locking.html#ADVISORY-LOCKS)),

[...]

PR #3389, opened 2022-02-05
Switch Postgresql to use transaction scoped advistory lock

jmotex:
https://github.com/flyway/flyway/issues/2895


jmotex:
After investigating and considering this, I switched it to transaction scoped advisory locks and tested locally with pgbouncer transaction pooling, session pooling and direct backend.  All three work the same and it works nicely with both per-migration lock and entire migration group lock.  The reported error could not be reproduced.  If somebody could take a look at the PR, it would be appreciated.

jmotex:
Found an issue, will work on that and resubmit when ready.

PR #3394, opened 2022-02-09
Support for PgBouncer transaction pooling

jmotex:
Hello Flyway Community!

This PR is for Flyway [Support for PgBouncer transaction pooling](https://github.com/flyway/flyway/issues/2895).  It's a small change only inside the Postgresql support code, so hopefully this will be okay.

1) The session scoped advisory lock has been switched for a transaction scoped advisory lock.

2) To make this work with non-transactional statements or migrations in the mix, Postgresql has been switched to use the dual connection support already built into Flyway (one for migrations and a main one for schema history table/lock control).  The advisory lock transaction lives on the main connection, while the migrations happen without the advisory lock transaction overshadowing them.  I couldn't find anything issues or commits in GitHub to suggest this was disabled to fix a bug, so am hoping it was just single connection as it was all that was needed until now.

This was tested with transactional & non-transactional SQL statements/migrations in the mix.  As well as non-transactional Java migrations in the mix on both session pool and transaction pooling modes of pgBouncer.  Concurrent migrations up to 4 in parallel was tested.

I can't think of any documentation changes from what I read, as all features/configs remain the same and migration transaction behavior also stays the same.


gmcquillan:
This is a major sticking point in our Infrastructure. 

Flyway needs to support the restricted transactional scope. PgBouncer is vital for proper connection coordination in many (maybe a majority) of Postgresql deployments, and using `session` pooling mode mostly defeats the purpose of having such a proxy (it cannot multiplex connections in this mode).

Importantly, `transaction` mode PgBouncer works fine for everything besides migrations. Can we get some eyes and attention on this?

gmcquillan:
@DoodleBobBuffPants and the Flyway Crew, What's the best way to get this feature onto the review pipeline?

sharkymcdongles:
It would be amazing to get this added in.

sharkymcdongles:
Seems mergeable! 👯 

Maybe we can get this in the next release? 🙏

sharkymcdongles:
We took the code here and built it into the latest released flyway to test. We then had 2 pods of the same service run against a bouncer in transaction pooling mode. I set surge to boot both simultaneously. It seems to finish successfully without issues. Great work!

sharkymcdongles:
@DoodleBobBuffPants I hope you're having a wondrous day and are bored and want to review this. <3

DoodleBobBuffPants:
Thanks for all the interest in this!

We've definitely got this on our radar, but are currently busy preparing for Flyway V9, however we'll revisit this as soon as we can!

sharkymcdongles:
> Thanks for all the interest in this!
> 
> We've definitely got this on our radar, but are currently busy preparing for Flyway V9, however we'll revisit this as soon as we can!

Any chance we can merge this in since v9 is now out?

DoodleBobBuffPants:
According to the [documentation](https://www.postgresql.org/docs/current/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY) for `CREATE INDEX CONCURRENTLY`, it seems this statement waits for active transactions to close before executing, so this PR cannot be merged as it prevents this statement from completing

DoodleBobBuffPants:
We've decided to make the locking mode configurable, you can find the documentation that is pending [here](https://github.com/flyway/flywaydb.org/pull/475). It'll default to being a transactional lock, but you can change the parameter to use a session-level lock for this edge case

sharkymcdongles:
To be clear I think you mean it will default to session level and then use transaction level to satisfy this edge case right? Because original behavior was session level. Just want to be clear. Also thanks so much for the response on this! <3

DoodleBobBuffPants:
While session-level locking was the original behaviour, we decided to make transactional locking the default behaviour now, since `CREATE INDEX CONCURRENTLY` is the outlier as far as we can tell

sharkymcdongles:
Okay thanks so much for the clarification! Eagerly awaiting 9.0.5!

ThommyH:
Using the new locks and transaction pooling comes with another problem: 
If you apply migrations and not have the schema set inside the migration files, it can happen, that the migration is applied against the public schema instead. That happenes, e.g. if you have `server_reset_query_always=1` set in pgbouncer config or have bad luck (your connections are shared, therefore the connection you use may has a state (wrong or no search path!). 
To mitigate this, the only way we found is to prepend every migration file with `SET SEARCH_PATH TO ${flyway:defaultSchema};`. But this seems impractical. 

Does anyone have a better solution for this? Unfortunately init-sql runs outside the transaction of a migration and therefore gets lost.

DoodleBobBuffPants:
@ThommyH Could you try a `beforeEachMigrate` [callback](https://flywaydb.org/documentation/concepts/callbacks#beforeEachMigrate)? These are supposed to run in the same transaction as each migration, so may help

ThommyH:
@DoodleBobBuffPants that worked for us, thank you!

PR #3416, opened 2022-03-25
Remove unnecessary empty log-line

pjwalstrom:


DoodleBobBuffPants:
while we appreciate this PR, we will not be merging it unless there is a reason behind it

pjwalstrom:
Thank you for the comment, highly appreciated. The only reason is that it takes up unnecessary space in e.g. Kibana. If there is an empty line, I assume there is a reason for it and someone tries to tell me something. In this case, I can not see the reason
flyway_empty_logline

PR #3432, opened 2022-04-13
hsqldb adjustBlockDepth while/loop/repeat

wangbfqq:
update hsqldb FUNCTION/PROCEDURE/TRIGGER block error


wangbfqq:
have agreed to the CLA

PR #3438, opened 2022-05-02
Snowflake - show procedures - exclude built-in

iapethus:
Snowflake recently introduced a change that shows also built-in procedures for `SHOW PROCEDURES`, which breaks the clean action.
https://community.snowflake.com/s/article/SHOW-PROCEDURES-Command-Output-Includes-Both-User-created-and-Built-in-Stored-Procedures

The proposed patch limits the output to only user procedures.


jwnorma:
The above change was recently deployed as part of Snowflake Release 6.37, and broke the clean action in flyway. 

Is it possible to merge this PR to fix? 

Thanks!

vgw-chriskruger:
Indeed this would be very nice to have.

PR #3456, opened 2022-05-31
update postgresql to 42.3.6

FunkyMuse:


DoodleBobBuffPants:
Thanks for the PR, however we won't be merging it as we have our own update mechanism and on first look it wouldn't resolve the related issue

Issue #3342, opened 2021-12-09
Broken Automatic-Module-Name for flyway-mysql, flyway-sqlserver


A248:
JPMS module names must be unique. It seems that, in the creation of flyway-mysql, the module name for flyway-mysql was copied and pasted from the core flyway artifact. So, the flyway-mysql module has the wrong module name. This causes usage with JPMS to be impossible.

https://github.com/flyway/flyway/blob/142a1067ca85b2245905bebf4cd8d906e22356e2/flyway-mysql/pom.xml#L151

Also, I don't use SQL Server, but it seems the same mistake was made there:

https://github.com/flyway/flyway/blob/142a1067ca85b2245905bebf4cd8d906e22356e2/flyway-sqlserver/pom.xml#L162

###### Which version and edition of Flyway are you using?
Flyway 8.2.1, Community Edition

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin)
Java API

###### Which database are you using? (Type & version)
Not applicable

###### Which operating system are you using?
Not applicable

###### What did you do? (Please include the content causing the issue, any relevant configuration settings, the SQL statement(s) that failed (if any), and the command you ran)
Run using the Java Module System (JPMS)

###### What did you expect to see?
Unique module names.

###### What did you see instead?
Ambiguous module names. The `flyway-mysql` artifact has the same module name as the core flyway artifact.

A248:
The same mistake was repeated for the firebird module:
https://github.com/flyway/flyway/blob/553d1c0fb6bcb111b95b9b00b6199d33e4c4c319/flyway-firebird/pom.xml#L128

The issue is these module names are all the same. They must be distinct in order to be valid modules.

The bug isn't very difficult to fix. Choose distinct module names: for example, `org.flywaydb.mysql` for the MySQL plugin.

agavrilov76:
I cannot upgrade to version 8.X due to this issue. Made a PR with a possible fix to speed up the process:
https://github.com/flyway/flyway/pull/3462

Note that the package name `org.flywaydb.database` is used across  the 'database' modules. So depending on db multiple modules at the same time wouldn't work due a package split.

cowwoc:
Come on Redgate... Where is the transparency? When are you going to fix this critical bug?!

Djaytan:
Hello,

Any news on it? This is a blocking issue when migrating to JPMS while relying on Flyway.

[...]

PR #3462, opened 2022-06-12
Fix automatic module name

agavrilov76:
Make sure database and communitiy modules have a differnt automatic module name then the core module
Fixes https://github.com/flyway/flyway/issues/3342

### DIFF
--- a/flyway-community-db-support/pom.xml
+++ b/flyway-community-db-support/pom.xml
@@ -98,8 +98,8 @@
                 <artifactId>maven-bundle-plugin</artifactId>
                 <configuration>
                     <instructions>
-                        <Automatic-Module-Name>org.flywaydb.core</Automatic-Module-Name>
-                        <Bundle-SymbolicName>org.flywaydb.core</Bundle-SymbolicName>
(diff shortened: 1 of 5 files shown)


A248 (approved):
This PR will successfully solve the module name problem for most deployments. I'd suggest that it be merged assuming it passes the proprietary test suite.

As mentioned in the linked issue, this PR won't solve the case where dependencies exist on more than one additional module. For example, using flyway-mysql and flyway-gcp-spanner at the same time is impossible. However, this PR is nevertheless worth merging, as it will un-block many, if not most, use cases until a further solution is developed. To fully solve the issue, some refactoring is necessary to resolve split packages.

agavrilov76:
Can someone please merge it? We cannot upgrade to Flyway 8+ due to this issue and considering stop using Flyway

mrotteveel:
This change is probably not sufficient, as these modules also have split packages with each other (specifically, the package `org.flywaydb.database`), and therefor would not be loadable as a module if more than one of these is on the module-path.

PR #3495, opened 2022-08-08
Kinetica database support

luissemantic:
Please note Kinetica is an up and coming relational database  that was not supported in Flyway.  Please see https://www.kinetica.com/ and also https://docs.kinetica.com/7.1/.  
Flyway would normally issue a commit statement after a script is run or rollback the transaction if the script fails.  However, one of the challenges for supporting Kinetica is that Kinetica does not have a concept of transactions. Commit and rollback commands don't exist and some classes in Flyway had to be extended and overridden to avoid issuing commits and rollbacks.


piers-williams:

PR #3510, opened 2022-08-26
closed if statement on same line

fiagaj:


DoodleBobBuffPants:
Thanks for the PR, however we won't be merging

Please see the comment in [this](https://github.com/flyway/flyway/issues/2908) issue as to why

PR #3543, opened 2022-10-20
Fix typo in error message

odin568:


Barry-RG:
This gramma is correct in context and the additional spacing is required for our pipeline. 
However, thank you for your PR.

PR #3554, opened 2022-10-29
chore(build): update maven wrapper to 3.8.6

reneleonhardt:


DoodleBobBuffPants:
Could you provide some context behind needing this change?

We likely won't accept this unless there's a larger need, so closing for now

PR #3577, opened 2022-11-30
Community database support for DB2 for z/OS

gahoekstraibm:
As part of a software solution we are maintaining, we introduced Flyway to manage our solution's database schemas. However, we needed to support both DB2 LUW and DB2 for z/OS, so we added support for DB2 for z/OS as a community extension to Flyway (DB2z). This extension supports regular statements to manage database objects as well as callable statements to e.g. run some database utility like REORG or CHECK DATA. We also collaborated with one of our customers that was considering to use this new DB2z support independently for other projects.

To properly support DB2 for z/OS, we needed to make a few additions to the Flyway core though:

1. In JDBCTemplate, we added a method to support parameterized utility calls as part of Flyway migrations (with proper error handling). DB2z requires to use a prepared statement for such parameterized calls, and JDBCTemplate was missing a method to do such calls with regular error handling.
2. We needed to add two new Flyway properties, specific for DB2z . One to provide the database name (which is a separate concept in DB2z, similar to a schema, but then used to hold storage related objects like table spaces etc), and another one to provide the SQLID to be used to create database objects (which may differ from the user used to connect to the database and may also differ from the schema name). Or is there a more decoupled way to add configuration properties for community database support?
3. To distinguish between DB2 LUW and DB2z, we needed to add a condition on database product version to the existing DB2 support. If the product version starts with "DSN", we are talking to a DB2z database, while if it starts with something else, it is DB2 LUW.

After using our customized Flyway in our large agile project for more than a year with DB2 for z/OS and also after our customer has tested Flyway z/OS support independently on other projects, we thought it would be a good thing to contribute this support for DB2 for z/OS back to the Flyway Community (as "compatible" database engine).

We are happy to answer any questions or process any findings or improvements!

Geert Hoekstra
IBM Netherlands


piers-williams:

PR #3588, opened 2022-12-22
fix Gradle plugin configuration syntax in documentation

halcyon22:
I encountered errors using the `postgresqlTransactionalLock` example using Gradle 6.8.3 and 7.4, so this corrects the syntax for all similar examples that I could find.

In "documentation/Flyway CLI and API/Configuration/Parameters/PostgreSQL Transactional Lock.md", without the change to line 42, Gradle is unable to compile the build script, reporting:

> You tried to use a map entry for an index operation, this is not allowed. Maybe something should be set in parentheses or a comma is missing?

Without the change to line 43, the `flywayMigrate` task fails with:

> class java.lang.Boolean cannot be cast to class java.lang.String (java.lang.Boolean and java.lang.String are in module java.base of loader 'bootstrap')


DoodleBobBuffPants:
Thanks for the PR!

Issue #3598, opened 2023-01-08
Documentation - Incorrect tar syntax


burgwyn:
###### Which version and edition of Flyway are you using?

Documentation site - https://documentation.red-gate.com/fd/command-line-184127404.html

The Linux CLI installation command has the incorrect syntax for tar, should be `tar -xvz`

tjs145:
This should be fixed now. Thanks

PR #3599, opened 2023-01-08
#3598 - Correct Linux command

burgwyn:
Addresses #3598 Corrects the syntax for the `tar` step


tjs145:
Thanks for letting us know about this. We've run the changes to our docs internally and the changes should now be live, so we can close this.

PR #3625, opened 2023-02-18
Update documentation to match code

JacksonBailey:
See `LogFactory.java` where it checks for `slf4j`, not `slf4j2`.


JacksonBailey:
Using `slf4j2` as the documentation says results in a class not found error. See [`LogFactory`](https://github.com/flyway/flyway/blob/main/flyway-core/src/main/java/org/flywaydb/core/api/logging/LogFactory.java#L105) for context.

JacksonBailey:
The documentation has been corrected (or removed) at some point. A search for `slf4j2` shows nothing. This is no longer relevant.

PR #3631, opened 2023-03-02
Prevent unnecessary statements in RedshiftDatabaseType

dreis2211:
Hi,

I noticed that while running on a normal Postgres database, there are some unnecessary queries for the version when checking the `RedshiftDatabaseType`.

image

This isn't a huge deal (it only takes 2% of the complete Flyway interaction for me). Overall only ~0.1% on startup, but nonetheless I think this is an avoidable interaction with the database when running on a classic Postgres instance.

Cheers,
Christoph


Barry-RG:
This has been included in our pipeline for the next release

dreis2211:
I don't see the commits really in main. Is this really merged properly?

PR #3644, opened 2023-03-27
Reduce allocations from class loading through FeatureDetector

dreis2211:
Hi,

I've noticed a couple of avoidable allocations and class lookups caused by `FeatureDetector`.
image

The idea of this PR is to check for the protocol first and only afterwards do the relatively "expensive" class loading.

Cheers,
Christoph


flyway-core/src/main/java/org/flywaydb/core/internal/scanner/Scanner.java
```diff
@@ -86,7 +84,7 @@ public Scanner(
 
 
             } else if (location.isAwsS3()) {
-                if (aws) {
+                if (detector.isAwsAvailable()) {
```
dreis2211, line 87: Is cached anyhow - avoids overhead if there are no AWS locations.....

Barry-RG:
This has been brought into our pipeline and will be included in a future release. Thank you for your contribution.

PR #3656, opened 2023-04-18
QuestDB Support

davidcooke2:
QuestDB is compatible with the postgres wire protocol, however lacks many of the features used by flyway such as locks, as with the SQLite integration, unsupported operations are no-ops. This also includes schema support, when using the postgres wire protocol, everything is on the `qdb` schema, this is implemented by ignoring schema changes and always returning the same schema.


Palmr:
@Barry-RG As someone else looking for QuestDB support in Flyway, is this going to get merged?

It seems new versions of `flyway-community-db-support` get published to maven, last one 11 days ago, yet seemingly all the PR's tagged with `database support` seem to be left open and never merged?

Are people not supposed to use `flyway-community-db-support` and instead pull from the various branches they need and build it themselves?

piers-williams:

PR #3674, opened 2023-05-17
Do not use MySQL named locks on Galera clusters

jasperkrijgsman:
Since MariaDB MDEV-30473 calling GET_LOCK on a (Galera) cluster is no longer allowed. Before this change the call was not replicated and only executed locally.

FlyWay now triggers the error "This version of MariaDB doesn't yet support 'GET_LOCK in cluster (WSREP_ON=ON)'" on Galera clusters.

This change checks for WSREP_ON=ON and disables the use of named locks and uses a table lock instead.

This functions in the same way as the check on Percona XtraDB Cluster with Strict Mode.


Barry-RG:
Thank you for your contribution, this will be included in our pipeline.

grooverdan:
@jasperkrijgsman thanks. FYI upstream bug report [MDEV-31325](https://jira.mariadb.org/browse/MDEV-31325).

PR #3678, opened 2023-05-26
refactor: Remove whitespace

birariro:
Avoid laborious scrolling


Barry-RG:
Unfortunately this required by our development process in managing both our open source and paid for otherings. We are making progress towards provide Flyway as open source without the necessity for these white spaces but until then these must remain.

PR #3697, opened 2023-06-29
Replace Jackson usage in ClassicConfiguration with Gson

gastaldi:
- Jackson is optional and therefore compilation fails while building a native image. Because Flyway already depends on Gson, it's better to use that instead

See https://github.com/quarkusio/quarkus/pull/34262#issuecomment-1612559070 for the error that's thrown in 9.20.0


Barry-RG:
@gastaldi We are looking to move away form GSON to Jackson as we use Jackson for other mapping as well as Json. As a result this PR is not compatible with that roadmap. Please bare with us.

PR #3731, opened 2023-08-22
Make SLF4J detection also work with SLF4J 2.0

### LINKED ISSUE #3730: SLF4J Autodetection doesn't work with SLF4J 2.0

9.21.0 Community

Yes.

Java API

PostgreSQL 15.3

Kubuntu 22.04

I ran Flyway in an integration test project with logback-classic 1.4.8 and slf4j-api 2.0.7 on the classpath.

Flyway logging via SLF4J

Flyway logging to stderr

#### Cause

I debugged through the issue and found that the error seems to be [here](https://github.com/flyway/flyway/blame/84b4dfbf03156561d632164eb3d77c2ed2154380/flyway-core/src/main/java/org/flywaydb/core/internal/util/FeatureDetector.java#L64-L67) – while the detection of the newer SLF4J version works fine (L.63), the result is afterwards set back to `false`, because the new implementation isn't returned by the `ServiceLoader` mechanism (empty stream -> `allMatch` is trivially true -> slf4j is marked as not available).

As Flyway uses the old SLF4J 1.7 API, and the check introduced in 9.16.2 only seems to be to exclude the slf4j-nop binding added to flyway-commandline, I'd suggest doing the check only for the old API as well.

hho:
#### Fixes #3730 :

With SLF4J >=1.8, the ServiceLoader mechanism doesn't return any loggers. Therefore, the `allMatch` evaluates to true, disabling SLF4J.

This change therefore constrains the ServiceLoader check to the old version of SLF4J.

* It shouldn't change the behavior for users of old SLF4J (such as flyway-commandline), as the ServiceLoader is still only called when an SLF4J binding is present
* It does enable the auto-detection of SLF4J 1.8 and and higher again, which was broken since Flyway 9.16.2, when the ServiceLoader check was introduced.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/util/FeatureDetector.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/util/FeatureDetector.java
@@ -58,13 +58,11 @@ public final class FeatureDetector {
             // provide any implementation, causing SLF4J to drop what we want to be console output on the floor.
             // Versions up to 1.7 have a StaticLoggerBinder
             slf4jAvailable = ClassUtils.isPresent("org.slf4j.Logger", classLoader)
-                    && ClassUtils.isPresent("org.slf4j.impl.StaticLoggerBinder", classLoader);
+                    && ClassUtils.isPresent("org.slf4j.impl.StaticLoggerBinder", classLoader)
+                    && !StreamSupport.stream(ServiceLoader.load(org.slf4j.Logger.class, classLoader)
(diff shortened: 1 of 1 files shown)


hho:
Sorry, regarding the failed checks: I don't understand how my change (no new code, no removed code, only a move) can fail compilation in a class I haven't touched?

Barry-RG:
Thank is understand able. I am aware of the issues with the build checks here and I am going to elevate this PR to join our pipeline.

PR #3764, opened 2023-10-19
Fix #3760 Avoid downloading migrations from S3 unless they will actually be run, when possible

### LINKED ISSUE #3760: Large migrations stored in S3 are downloaded, even when they aren't run.
9.22.3 Community

Maven plugin (9.21.1)

MariaDB 10.6

Windows (10/11) & Linux (Amazon Linux 2)

I have a 1.3GB migration stored in S3.
I ran flyway repair & flyway migrate against an empty DB, and against a DB where flyway migrate has already been run.

In the case where the DB is empty, I expect the large migration to not be downloaded when running flyway repair, and then to be downloaded once from S3, then applied, when running flyway migrate

In the case where the DB is populated, and the large migration has already run, I expect flyway to not download the migration at all for either repair or migrate operations.

In the case where the DB is empty, the migration is downloaded multiple times. Once when running repair and once when running migrate.

In the case where the DB is populated, the migration is downloaded despite already having been run.

In both cases, due to the size of the migration file, multiple minutes are spent redownloading the file when it doesn't seem like it should be necessary to do so if the migration isn't going to run.

OneGeek:
Addresses flyway/flyway#3760

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/api/resource/LoadableResource.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/api/resource/LoadableResource.java
@@ -15,7 +15,15 @@
  */
 package org.flywaydb.core.api.resource;
 
+import java.io.BufferedReader;
+import java.io.IOException;
 import java.io.Reader;
+import java.nio.charset.StandardCharsets;
+import java.util.zip.CRC32;
+
+import org.flywaydb.core.api.FlywayException;
+import org.flywaydb.core.internal.util.BomFilter;
+import org.flywaydb.core.internal.util.IOUtils;
 
 /**
  * A loadable resource.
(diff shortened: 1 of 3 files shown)


OneGeek:
As currently implemented this retrieves the crc32 checksum that can be stored in an S3 object upon upload. If it wasn't supplied, or fails to retrieve the checksum for any reason, it falls back to the current behavior of downloading the file and computing the checksum locally.

Note that in order for S3 to have the CRC32 available, it needs to have been included (or marked for computation) at the time of the upload to S3, otherwise it'll be null. S3 does have an etag field that's a hash of the contents which could be grabbed and then checksummed with crc32 in order to make things work without needing to do anything in particular when uploading to S3, but then the migration checksum when stored on S3 would differ from the same migration stored locally. Depending on exactly the goals of the flyway project, that (or an option to choose which behavior) may or may not be acceptable, but I don't have enough context to make that call, so I chose the conservative approach.

I expect to need to update docs before the PR is accepted, but wanted to confirm this change is acceptable before putting in that effort.

OneGeek:
Is this something that might be incorporated into flyway in the near future?

piers-williams (changes requested):
Apologies it has taken so long.

It is a nice idea, though there are some issues at present. Repeatable migrations have the checksum calculated *after* placeholder replacement and so will need to always have the full download.

OneGeek:
Thanks for taking a look in any case!

Specific migrations can be excluded from this behavior by not setting their checksum value (since it's not an automatic thing on the S3 side) and this PR already handles them being missing by downloading anyway.

Alternatively, is it common to have migrations with placeholders that are multiple GB in size? If that's sufficiently rare, having a configurable threshold for whether or not to care about the checksum might be reasonable.

skrcode:
@OneGeek — I ran JAIPilot Cloud against this exact head (dff09a2). It produced a small AI-generated cleanup draft limited to the three production files already changed here: https://github.com/skrcode/flyway/pull/2

A temporary 7-test checksum/S3 harness passed before and after, then was omitted because retaining its required POM dependency would exceed PR scope. The final production-only diff passed the clean build. This is a review aid only; nothing is auto-merged.

PR #3791, opened 2023-11-08
Fix #3786 Postgresql collations not dropped on clean

### LINKED ISSUE #3786: PostgreSQL Collations are not cleaned up on clean
We currently use version 9.22.0

I also tried it on 10.0.0 (i used the command line since flyway is currently broken with spring boot) but still same issue 

We use the Java Api together with spring boot

Postgresql 13

MacOS / Debian

I first created a migration which creates a collation.
```sql
CREATE COLLATION german (provider = icu, locale = 'de-DE');
```
And applied it

Then I ran flyway with these command line arguments:
`clean -url="jdbc:postgresql://localhost:5432/test" -user=postgres -cleanDisabled=false`

I expected that the collations are cleaned up after the clean command was run.

The collations where still persisted in the schema

Mihitoko:
This PR fixes #3786

We introduce a new query here that is executed when a PSQL schema is cleaned which cleans up all collations in that schema that don't depend on an extension.


piers-williams:
Thank you for your contribution, this has been incorporated into Flyway 10.22.0 released yesterday

Issue #2438, opened 2019-07-10
Redshift - Support for ALTER TABLE ALTER COLUMN


anjo-swe:
###### Which version and edition of Flyway are you using?

5.2.4

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin)

Command line

###### Which database are you using (type & version)?

Redshift

###### Which operating system are you using?

Linux

###### What did you do?

Ran a migration to update a table inside Redshift

```
ALTER TABLE lorem.my_table_name ALTER COLUMN type type varchar(30); 
```

###### What did you expect to see?

* Table should be altered
* Statement shouldn't be executed inside a transaction

###### What did you see instead?

```
ERROR: ALTER TABLE ALTER COLUMN cannot run inside a transaction block
```

##### References

https://docs.aws.amazon.com/redshift/latest/dg/r_ALTER_TABLE.html
> ALTER COLUMN column_name TYPE new_data_type
> ...
> * You can't alter columns inside a multi-statement block (BEGIN...END).

https://github.com/flyway/flyway/blob/master/flyway-core/src/main/java/org/flywaydb/core/internal/database/redshift/RedshiftParser.java#L45
* Need to check for `ALTER TABLE ALTER COLUMN` as well

Related to: https://github.com/flyway/flyway/issues/1879

MikielAgutu:
Thanks for the bug report, this is fixed and will be released in v6.

ajayl-tubi:
@anjo-swe Thanks for the fix! @MikielAgutu When is the v6 planned to release? As our team needs this fix soon.

radifalco:
I'm still seeing this. I get this error in a migration file. 

```
[Amazon](500310) Invalid operation: ALTER TABLE ALTER COLUMN cannot run inside a transaction block;
```

PR #3815, opened 2023-12-21
Redshift add support for `IDENTITY PROVIDER`

jon-ruckwood:
Redshift cannot execute [identity provider][0] statements from within a transaction block.

This is similar to the issue and fix in #2438.

[0]: https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_IDENTITY_PROVIDER.html


piers-williams:
Is this a change you are still interested in us taking a look at?

PR #3850, opened 2024-02-22
Update the Flyway plugin for YugabyteDB

Sfurti-yb:
Problems:
* The current implementation does not use jdbc:yugabytedb:// for the connections.
* If user is using Flyway baseline, the baseline has default implementation of putting DDL and DML in a single transaction which YB will roll back the DML because of the catalog version changes.
* For YugabyteDB v2.20 all the commands fail with the error: ERROR: advisory locks are not yet implemented which has not been an issue for previous versions and not errored out.
Solutions:
* Added jdbc:yugabytedb: protocol in YugabyteDBDatabaseType.java
* Changed the implementation of getRawCreateScript() so that DDLs and DMLs are not in a single transaction
* For the pg_advisory_lock issue, setting yb_silence_advisory_locks_not_supported_error=on where connection is create and returning that connection

Tests:
The tests for these changes are in the repository: https://github.com/yugabyte/flyway-tests

### DIFF
--- a/flyway-community-db-support/flyway-database-yugabytedb/pom.xml
+++ b/flyway-community-db-support/flyway-database-yugabytedb/pom.xml
@@ -39,6 +39,11 @@
             <artifactId>flyway-database-postgresql</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <scope>compile</scope>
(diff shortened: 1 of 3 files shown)


Sfurti-yb:
Hi @Barry-RG I have raised a PR Here: https://github.com/flyway/flyway-community-db-support/pull/27

Barry-RG:
Closing as this has been ported and merged in the community DB Support repository

PR #3879, opened 2024-04-21
Gradle: Update Gradle Plugin extension to use properties

Nava2:
This updates the flyway gradle plugin to provide `Property<>` values on the extension rather than using bare types. This is incredibly important for working with Gradle such that properties remain lazy. This allows values to be configured from different locations rather than relying on being available immediately.

Drive-by: This bumps the version of Gradle from 6.1.1 (Jan 2020) to 6.9.4 (Feb 2023). The version in use is _wildly_ out of date. 

All changes within this PR are backwards compatible with Version 6.9.4.


WilliamL-RG (changes requested):
Apologies for the time since this PR was created. I have found some issues with the PR that are preventing Flyway working as expected with Gradle, so they will need to be addressed.

PR #3912, opened 2024-06-19
Update Getting Started link in README.md

odolbeau:


piers-williams:
Thank you for reporting this - it has now been fixed

PR #3968, opened 2024-10-21
Avoid generating FlywaySqlScriptException message dynamically

gsmet:
This is suboptimal as we end up generating the same string over and over again and this particular `getMessage()` requires class loader calls (via `resource.getAbsolutePathOnDisk()`).

In Quarkus, we actually call `getMessage()` after the class loader is closed in some cases which causes some issues.

I think this is good practice anyway as better avoid calling the class loader multiple times when an error occurs.

See https://github.com/quarkusio/quarkus/issues/43926 for the background of this change.


gsmet:
Ping? We at Quarkus would really like to get this in as we can't properly get feedback to users in some cases. Thanks!

gsmet:
This can be closed as the issue has been fixed in `11.3.0`

PR #4010, opened 2025-01-13
[snowflake] fix parser

mlorek:
snowflake parser fails to parse the following command:
```
COMMENT IF EXISTS ON COLUMN sch.tbl.col IS 'text';
```


WilliamL-RG:
Thank you for your contribution, this has been added into Flyway in 11.2.0

Issue #4014, opened 2025-01-25
Mysql performance_schema exception is com.mysql.cj.exceptions.CJException and not SQLException


MilovdZee:
mysql-connector-j-9.1.0
flyway 11.2.0

During the check for access to the performance_schema, flyway relies on a specific exception to be thrown if the access is absent. Issue seems to be that the exception is not "java.sql.SQLException" but in seems to be "com.mysql.cj.exceptions.CJException".
That exception is not catched and so the migration fails.

PR #4016, opened 2025-01-25
issue #4014: crash when the mysql user does not have read rights

MilovdZee:
on performance_schema.user_variables_by_thread.

Now just catching the exception.


MilovdZee:
I know that is generally bad practice to catch Exception but in this case it won't hurt anything. The code is just to check if there is access to performance_schema.user_variables_by_thread and returning false is always fine.

Issue #3871, opened 2024-04-12
Oracle: Flyway parsing bug: unable to decrease block depth below 0 


YevheniiaSilbernagl:
###### Which version and edition of Flyway are you using?
```
<flyway.version>10.11.0</flyway.version>
<dependency>
            <groupId>org.flywaydb</groupId>
            <artifactId>flyway-core</artifactId>
            <version>${flyway.version}</version>
        </dependency>
        <dependency>
            <groupId>org.flywaydb</groupId>
            <artifactId>flyway-database-oracle</artifactId>
            <version>${flyway.version}</version>
            <scope>test</scope>
        </dependency>
```
###### If this is not the latest version, can you reproduce the issue with the latest one as well? (Many bugs are fixed in newer releases and upgrading will often resolve the issue)
it's the last version that I can see in https://mvnrepository.com/

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin)
It's integration tests in a springboot application

###### Which database are you using? (Type & version)
**Oracle 21.0.0.0.0** from Docker image: **gvenzl/oracle-xe:21.3.0-slim**

###### Which operating system are you using?
I'm running tests on Windows, Oracle is running on Linux

###### What did you do? (Please include the content causing the issue, any relevant configuration settings, the SQL statement(s) that failed (if any), and the command you ran)
migraton query that it's fialing on 
```
CREATE OR REPLACE PACKAGE AFCSYS.ACCESS_RIGHTS AS
    PROCEDURE set_context (roleId IN NUMBER, userId IN NUMBER);
END AFCSYS.ACCESS_RIGHTS;

CREATE OR REPLACE PACKAGE BODY AFCSYS.ACCESS_RIGHTS AS
    PROCEDURE set_context (roleId IN NUMBER, userId IN NUMBER) IS
    BEGIN
       NULL;
    END set_context;
END AFCSYS.ACCESS_RIGHTS;
```

###### What did you expect to see?
I was expecting it to migrate with no errors

###### What did you see instead?
```
DbValidate:37 Successfully validated 6 migrations
JdbcTableSchemaHistory:37 Creating Schema History table "SYSTEM"."flyway_schema_history" with baseline ...
DbBaseline:37 Successfully baselined schema with version: 1
DbMigrate:37 Current version of schema "SYSTEM": 1
DbMigrate:37 Migrating schema "SYSTEM" to version "2 - table space"
[...]

PR #4073, opened 2025-05-01
Fix OracleParser block-depth underflow (#3871)

W25X80:
Fix OracleParser underflow in PACKAGE BODY parsing

Handle END <pkg> specially and guard generic END with blockDepth>0 to prevent the “unable to decrease block depth below 0” error.

### DIFF
--- a/flyway-database/flyway-database-oracle/src/main/java/org/flywaydb/database/oracle/OracleParser.java
+++ b/flyway-database/flyway-database-oracle/src/main/java/org/flywaydb/database/oracle/OracleParser.java
@@ -52,6 +52,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.regex.Pattern;
+import java.util.regex.Matcher;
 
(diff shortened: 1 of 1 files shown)


JasonLuo-Redgate:
Hi @W25X80 ,

Could you provide a sample SQL segment that currently causes a parsing issue, and which this PR is intended to fix?

W25X80:
Hi @JasonLuo-Redgate ,

```
CREATE OR REPLACE PACKAGE AFCSYS.ACCESS_RIGHTS AS
    PROCEDURE set_context (roleId IN NUMBER, userId IN NUMBER);
END AFCSYS.ACCESS_RIGHTS;

CREATE OR REPLACE PACKAGE BODY AFCSYS.ACCESS_RIGHTS AS
    PROCEDURE set_context (roleId IN NUMBER, userId IN NUMBER) IS
    BEGIN
       NULL;
    END set_context;
END AFCSYS.ACCESS_RIGHTS;
```

PR #4086, opened 2025-05-28
Fix #4063 Repeatable scripts validation fails when updating flyway v1…

### LINKED ISSUE #4063: Flyway validate fails for a placeholder based repeatable script with checksum mismatch
Flyway version:  v10.18.2 
Database Type: SQL Server
Database Version: Microsoft SQL Azure (RTM) - 12.0.2000.8 Feb 27 2025 21:10:52 Copyright (C) 2024 Microsoft Corporation
Client: Java API with Spring Boot 3.3
Is this Reproducible ? Yes (But, cannot write a test to use two different flyway versions)

Problem: 
After upgrading to Flyway from 10.10.0 to 10.18.2, we are seeing a flyway validation error due to checksum mismatch. 

`org.flywaydb.core.api.exception.FlywayValidateException: Validate failed: Migrations have failed validation\nMigration checksum mismatch for migration R__Login_History.sql\n-> Applied to database : 1749060336\n-> Resolved locally    : -428002767\nEither revert the changes to the migration, or run repair to update the schema history.\nNeed more flexibility with validation rules? Learn more: [https://rd.gt/3AbJUZE\\n](https://rd.gt/3AbJUZE%5C%5Cn) at org.flywaydb.core.Flyway.lambda$migrate$0(Flyway.java:180)\n at org.flywaydb.core.FlywayExecutor.execute(FlywayExecutor.java:212)\n at org.flywaydb.core.Flyway.migrate(Flyway.java:167)\n at org.springframework.boot.autoconfigure.flyway.FlywayMigrationInitializer.afterPropertiesSet(FlywayMigrationInitializer.java:66)\n at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.invokeInitMethods(AbstractAutowireCapableBeanFactory.java:1853)\n at` 

It is strange that, it is failing after a minor upgrade, given this file was never modified for a long time now.  (applied with Flyway v6.0.8)

We tried to dwell deeper into the problem, and figured, a change was made with flyway-10.17.x - 

[](https://github.com/flyway/flyway/blob/bc7d8a894f148a6d32808ecbbd1ca0374daeb9e3/flyway-core/src/main/java/org/flywaydb/core/internal/info/MigrationInfoImpl.java#L301)

Image

However, in other places of finding the `state` of the applied migrations to compare against resolved migrations - we are using this checksum comparison.

[](https://github.com/flyway/flyway/blob/bc7d8a894f148a6d32808ecbbd1ca0374daeb9e3/flyway-core/src/main/java/org/flywaydb/core/internal/resolver/ResolvedMigrationImpl.java#L132)

It appears, this is a broken compatibility problem. 

The only way out of it seems like, repair or modifying all the repeatable scripts, if we want to upgrade to this version of the flyway. 

If this is a bug, can you please handle it. ?

If not any suggestions are welcome.

RaviKalasapur:
An unmodified repeatable script (from flyway v7) fails to be applied when flyway in the project is updated from flyway v10.10 .0 to v10.18.2. 

Since, its a repeatable script, it has to be recognized as outdated and applied, instead of failing at validation.

### DIFF
--- a/flyway-core/src/main/java/org/flywaydb/core/internal/info/MigrationInfoImpl.java
+++ b/flyway-core/src/main/java/org/flywaydb/core/internal/info/MigrationInfoImpl.java
@@ -458,6 +458,14 @@ public class MigrationInfoImpl implements MigrationInfo {
         return result;
     }
 
+    /**
+     * Overriden to handle repeatable script checksum mismatch on migrate - https://github.com/flyway/flyway/issues/4063
+     */
+    @Override
+    public boolean isChecksumMatching() {
+        return this.resolvedMigration.checksumMatches(this.appliedMigration.getChecksum());
+    }
+
     @Override
     public Integer getResolvedChecksum() {
         return resolvedMigration == null ? null : resolvedMigration.getChecksum();


piers-williams:
Closing for same [reason as the issue](https://github.com/flyway/flyway/issues/4063)

Issue #3798, opened 2023-11-21
Gradle 8.4 deprecation warnings


SchlauFuchs:
###### Which version and edition of Flyway are you using?

9.21.1

###### Which client are you using? (Command-line, Java API, Maven plugin, Gradle plugin)

Gradle plugin

###### Which database are you using? (Type & version)

Informix V14.10

###### Which operating system are you using?

Windows

###### What did you do? (Please include the content causing the issue, any relevant configuration settings, the SQL statement(s) that failed (if any), and the command you ran)

running gradle flywayMigrate task

###### What did you expect to see?

no deprecation warnings in gradle output

###### What did you see instead?

```
> Task :flywayMigrate
The Project.getConvention() method has been deprecated. This is scheduled to be removed in Gradle 9.0. Consult the upgrading guide for further information: https://docs.gradle.org/8.4/userguide/upgrading_version_8.html#deprecated_access_to_conventions
The org.gradle.api.plugins.Convention type has been deprecated. This is scheduled to be removed in Gradle 9.0. Consult the upgrading guide for further information: https://docs.gradle.org/8.4/userguide/upgrading_version_8.html#deprecated_access_to_conventions
The org.gradle.api.plugins.JavaPluginConvention type has been deprecated. This is scheduled to be removed in Gradle 9.0. Consult the upgrading guide for further information: https://docs.gradle.org/8.4/userguide/upgrading_version_8.html#java_convention_deprecation
Flyway upgrade recommended: Informix 14.10 is newer than this version of Flyway and support has not been tested. The latest supported version of Informix is 12.10.
```

bvella:
I have encountered this warning as well. Comparing with the Jooq Plugin (https://github.com/etiennestuder/gradle-jooq-plugin/blob/main/src/main/groovy/nu/studer/gradle/jooq/JooqPlugin.java#L65), it seems that all you have to do to fix this is replace line 615 in AbstractFlywayTask with this:

```JavaPluginExtension plugin = (JavaPluginExtension)project.getExtensions().getByType(JavaPluginExtension.class)```

**Comment by mrclrchtr** [comment]
At first: Many thanks for providing this plugin.

But is there a maintainer here who can provide some feedback?

We actually want to keep our build free of warnings, but that's not possible because of this plugin 😕

**Comment by jjohannes** [comment]
With `Gradle 9 RC2` the plugin unfortunately no longer works:

```
FAILURE: Build failed with an exception.

* What went wrong:
[...]

PR #4089, opened 2025-05-30
Gradle Plugin: Migrate from deprecated 'convention' API

mokamoto12:
Updated Gradle plugin's build scripts to no longer use the deprecated `convention` property, switching to the `extensions` API instead.

close https://github.com/flyway/flyway/issues/3798


doctorpangloss:
this needs to be merged for gradle 9 support

WilliamL-RG:
Thanks for the contribution, this has been added into Flyway since 11.10.3

PR #4092, opened 2025-06-02
Gradle Plugin: Use task configuration avoidance API

mokamoto12:
This PR updates the Flyway Gradle plugin to use Gradle's [Task configuration avoidance API](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html).

This update is part of https://github.com/flyway/flyway/pull/3879, but I've created a separate PR for it as this change is useful on its own.


steven-sheehy:
Seems like it already uses register on main? https://github.com/flyway/flyway/blob/main/flyway-gradle-plugin/src/main/java/org/flywaydb/gradle/FlywayPlugin.java

WilliamL-RG:
Thanks for the contribution, this has been added into Flyway since 11.10.0

PR #4153, opened 2025-10-06
Fix Gradle dependency notation in Oracle DB docs

JohannesFreitag:
Hi,

As is already correctly stated in [the docs for the Gradle task](https://documentation.red-gate.com/flyway/reference/usage/gradle-task), flyway database plugins should nowadays need to be declared as Gradle buildscript dependencies.

The documentation for Oracle DB still contains some old snippets that reference the old way of declaring it as a project dependency, which does not work anymore.

I fixed that because it can be quite confusing when you stumble upon the Oracle docs and find contradictory information.

### DIFF
--- a/documentation/Reference/Database Driver Reference/Oracle Database.md
+++ b/documentation/Reference/Database Driver Reference/Oracle Database.md
@@ -58,16 +58,20 @@ Oracle support is a separate dependency for Flyway and will need to be added to 
 #### Redgate
 
 ```groovy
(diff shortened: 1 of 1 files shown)


alistair-RG:
@JohannesFreitag - thanks for this, we've included this in the docs now (part of Flyway 11.15.0 work)
https://documentation.red-gate.com/fd/oracle-database-277579323.html

PR #4187, opened 2025-12-10
Adding QuestDB to the list of supported systems on README.md

javier:
QuestDB has been supported for a while, but it is not listed on the README


alistair-RG:
We're currently making a few other changes to this page and will include this too - I'll close this when it's released

PR #4188, opened 2025-12-12
Class cannot be located on the disk when using instrumentation agents like jacoco

myroch:
Hello community
We have found the issue in our test using oracle db together with jacoco. Since 11.7.1 is failing this with NullPointerException. The issue was already reported to quarkus community: https://github.com/quarkusio/quarkus/issues/49566

This fix should handle this case.
Thank you very much for merging to mainstream.
Regards
myroch


JasonLuo-Redgate:
Hi @myroch ,

I just wanted to check something with you. Flyway uses the `getInstallDir` method in several places. Even if this particular case is fixed, I assume your application may still fail when running with `JaCoCo` if `getLocationOnDisk` returns `null`.

Do you have any suggestions on how we should handle this scenario? My concern is that fixing this single location may not fully resolve the issue.

myroch:
Hello @JasonLuo-Redgate 
actually this location is enough, at least for us. The tests work afterwards.

JasonLuo-Redgate:
Hi @myroch ,

Many thanks for the feedback. This issue has been fixed since `Flyway 12.1.1`.

Issue #4164, opened 2025-10-19
InsertRowLock requires the database to implement primary keys


jaccozilla:
I'm looking at adding support for locking in clickhouse https://github.com/flyway/flyway-community-db-support/issues/48. `InsertRowLock` is close to working, but because clickhouse doesn't support traditional primary keys I had to make some modifications to it. 

I'm allowing multiple locks to be written, but consistently selecting a winner by ordering on timestamp + version. This needs the `insertLockingRow` to also query the current owner.

I would like to refactor InsertRowLock a little and pull out an `AbstractInsertRowLock` allowing for subclasses to implement `insertLockingRow`.

This is what i'm currently using for clickhouse's `insertRowLock

```
private boolean insertLockingRow(String insertStatementTemplate, String currentLockOwnerStatementTemplate, String booleanTrue, String checksumValue) throws InterruptedException {

        String currentLockOwner = getCurrentLockOwner(currentLockOwnerStatementTemplate);
        if (tableLockString.equals(currentLockOwner)) {
            return true;
        }
        if (currentLockOwner != null) {
            // owned by another migration
            LOG.info("[{}] Unable to lock, already owned by {}",  tableLockString, currentLockOwner);
            return false;
        }

        String insertStatement = String.format(insertStatementTemplate.replace("?", "%s"),
                                               -100,
                                               "'" + tableLockString + "'",
                                               "'" + FLYWAY_LOCK_STRING + "'",
                                               "''",
                                               "''",
                                               checksumValue,
                                               "''",
                                               0,
                                               booleanTrue
                                              );

        // Insert the locking row - the primary key-ness of 'installed_rank' will prevent us having two
        Results results = jdbcTemplate.executeStatement(insertStatement);

        if (results.getException() != null) {
            return false;
        }

        if (currentLockOwnerStatementTemplate == null) {
            return true;
        }

        currentLockOwner = getCurrentLockOwner(currentLockOwnerStatementTemplate);
        if (tableLockString.equals(currentLockOwner)) {
            // double check we are still the current lock owner, in case we raced with another
            // insert and need version ordering to select the current owner
            currentLockOwner = getCurrentLockOwner(currentLockOwnerStatementTemplate);
            if (tableLockString.equals(currentLockOwner)) {
                // this row lock is the owner
                return true;
            }
        }

        // not the owner
        LOG.info("[{}] Attempted to lock, but already owned by {}",  tableLockString, currentLockOwner);
        return false;
    }

[...]

PR #4203, opened 2026-02-11
Add InsertRowLock#doLock method with lockStrategy predicate enabling alternative implementations for databases without primary key support

jaccozilla:
https://github.com/flyway/flyway/issues/4164

`RowLock` currently assumes the db has a primary key on `installed_rank`, for clickhouse this doesn't work. Small refactor to open up another `doLock` method which takes a predicate to do the locking which clickhouse can implement an alternative for


jaccozilla:
I didn't see anywhere to add tests? Can someone point me where to add those?

jaccozilla:
signed the cla

Issue #4191, opened 2026-01-02
Add Jackson 3.0 support


hurelhuyag:
Spring Boot 4.0 now depends on Jackson 3.0. Although the Jackson 2.x branch does not conflict with Jackson 3.0, it is more convenient to use a single Jackson library.

essicajayleigh-create:
![image](https://github.com/user-attachments/assets/f18d84e1-ad18-42c8-8c33-df34a253d5e0)

essicajayleigh-create:
[]()

ffray:
I'd like to contribute the required changes if possible.
Is this an option in this project?

ffray:
If possible, please have a look at the linked project via the issue link above.
It updates Jackson to version 3.x. 
Please give me some guidance whether such change would get accepted or not, so I can continue to work on this change or discard it.

ziqin:
I'd like to mention #4046 here. If flyway-core doesn't depend on Jackson anymore, there would be no 2.x vs 3.x problem.

ffray:
Uh, @WilliamL-RG, this is excellent news. Thanks for upgrading!

PR #4220, opened 2026-03-17
Standardize jackson-core in Docker images to pinned versions from Maven Central

jykae:
## Proposed changes

Flyway 12's Maven assembly bundles multiple jackson-core versions in `/flyway/lib/` because internal modules (`flyway-core`, `flyway-database-*`, `flyway-reports`) each resolve their own transitive jackson dependency. This triggers Trivy findings for the older versions.

## Changes

- **All three Dockerfile variants** (`base`, `alpine`, `azure`) that extract the flyway-commandline tarball now:
  - Delete all `jackson-core-*.jar` from the extracted distribution
  - Download the two required versions from Maven Central with SHA-256 verification
  - `com.fasterxml.jackson.core:jackson-core:2.21.1` (2.x API consumers)
  - `tools.jackson.core:jackson-core:3.1.0` (Flyway's own 3.x usage)

- **Versions and checksums are build ARGs** with defaults matching `pom.xml`, overridable at build time without code changes

```dockerfile
ARG JACKSON_CORE_V2=2.21.1
ARG JACKSON_CORE_V2_SHA256=1edd5f2e49dca5f8e4519957c24b7b3050bd1c7ee883920da33cff031ff1f7c0
ARG JACKSON_CORE_V3=3.1.0
ARG JACKSON_CORE_V3_SHA256=4dd383f96b51b9b9ac4b74bdf1c150df0100443e1da1fba480f57da09a2dcef7
```

Mono/oracle/redgate overlay Dockerfiles inherit from these base images — no changes needed there.

### DIFF
--- a/flyway-docker/dockerfiles/alpine/Dockerfile
+++ b/flyway-docker/dockerfiles/alpine/Dockerfile
@@ -1,7 +1,9 @@
 FROM bash:5 AS untar
 ARG FLYWAY_VERSION
 ARG JACKSON_CORE_V2=2.21.1
+ARG JACKSON_CORE_V2_SHA256=1edd5f2e49dca5f8e4519957c24b7b3050bd1c7ee883920da33cff031ff1f7c0
 ARG JACKSON_CORE_V3=3.1.0
+ARG JACKSON_CORE_V3_SHA256=4dd383f96b51b9b9ac4b74bdf1c150df0100443e1da1fba480f57da09a2dcef7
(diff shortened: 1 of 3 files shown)


jykae:
@copilot could you sign your contributor license agreement in PR comment

alistair-RG:
We believe this is now covered because of the work we did for #4191 that was released in Flyway v12.0.3

PR #4223, opened 2026-03-30
Make Flyway Gradle tasks compatible with configuration cache

### LINKED ISSUE #4107: Support Gradle configuration cache
The Flyway plugin does not support the Gradle configuration cache. With [Gradle 9.0](https://docs.gradle.org/9.0.0-rc-1/release-notes.html), the configuration cache becomes the default execution mode. Gradle states we should "prepare for a future where it becomes the only supported mode." There was a previous issue https://github.com/flyway/flyway/issues/3550 closed as not planned, but perhaps with this new importance being placed on the configuration cache by Gradle the maintainers can reconsider.

Workarounds like `tasks.withType<AbstractFlywayTask> { notCompatibleWithConfigurationCache("") }` also don't seem to work in my Gradle build.

- Task `:app:flywayMigrate` of type `org.flywaydb.gradle.task.FlywayMigrateTask`: invocation of 'Task.project' at execution time is unsupported.
  See https://docs.gradle.org/8.14/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution

ln-e:
Move all Task.getProject() calls from execution time to configuration time in AbstractFlywayTask to fix "Invocation of Task.project at execution time is unsupported" error when running with --configuration-cache.

closes #4107

### DIFF
--- a/flyway-plugins/flyway-gradle-plugin/src/main/java/org/flywaydb/gradle/task/AbstractFlywayTask.java
+++ b/flyway-plugins/flyway-gradle-plugin/src/main/java/org/flywaydb/gradle/task/AbstractFlywayTask.java
@@ -7,9 +7,9 @@
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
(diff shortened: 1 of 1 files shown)


alistair-RG:
This should go out in the next release (after Flyway 12.3.0)

PR #4252, opened 2026-06-24
Avoid NullPointerException in ClasspathSqlMigrationScanner when getResource(".") returns null

### LINKED ISSUE #4241: ClasspathSqlMigrationScanner.matchesPath NPEs under classloaders returning null for getResource(".")
Hi.
I'm building [flyway-mongodb extension](https://github.com/quarkusio/quarkus/issues/54224) for [quakus](https://quarkus.io/). Because I love quarkus, mongodb and flyway, and I want more people to use this combination seamlessly.

During implementation I faced an NullPointerException under quakus classloader which originates in ClasspathSqlMigrationScanner.

```
 2026-05-21 09:23:46,677 ERROR [io.quarkus.runtime.Application] (Quarkus Main Thread) Failed to start application: java.lang.RuntimeException: Failed to start quarkus
        at io.quarkus.runner.ApplicationImpl.doStart(Unknown Source)
        at io.quarkus.runtime.Application.start(Application.java:112)
        at io.quarkus.runtime.ApplicationLifecycleManager.run(ApplicationLifecycleManager.java:127)
        at io.quarkus.runtime.Quarkus.run(Quarkus.java:79)
        at io.quarkus.runtime.Quarkus.run(Quarkus.java:50)
        at io.quarkus.runtime.Quarkus.run(Quarkus.java:143)
        at io.quarkus.runner.GeneratedMain.main(Unknown Source)
  Caused by: org.flywaydb.core.api.FlywayException: Cannot invoke "java.net.URL.getPath()" because the return value of "java.lang.ClassLoader.getResource(String)" is null
        at org.flywaydb.nc.preparation.PreparationContext.getFromFuture(PreparationContext.java:159)
        at org.flywaydb.nc.preparation.PreparationContext.initialize(PreparationContext.java:76)
        at org.flywaydb.nc.preparation.PreparationContext.get(PreparationContext.java:148)
        at org.flywaydb.verb.migrate.MigrateVerbExtension.executeVerb(MigrateVerbExtension.java:70)
        at org.flywaydb.core.Flyway.migrate(Flyway.java:178)
        at io.quarkus.flyway.mongodb.runtime.FlywayMongodbRecorder.doStartActions(FlywayMongodbRecorder.java:137)
  Caused by: java.lang.NullPointerException: Cannot invoke "java.net.URL.getPath()" because the return value of "java.lang.ClassLoader.getResource(String)" is null
        at org.flywaydb.scanners.ClasspathSqlMigrationScanner.matchesPath(ClasspathSqlMigrationScanner.java:115)
        at org.flywaydb.scanners.BaseSqlMigrationScanner.lambda$scanFromFileSystem$0(BaseSqlMigrationScanner.java:93)
        at org.flywaydb.scanners.BaseSqlMigrationScanner.scanFromFileSystem(BaseSqlMigrationScanner.java:95)
        at org.flywaydb.scanners.ClasspathSqlMigrationScanner.scan(ClasspathSqlMigrationScanner.java:98)
        at org.flywaydb.nc.migration.MigrationScannerManager.lambda$scan$4(MigrationScannerManager.java:142)
        at org.flywaydb.nc.migration.MigrationScannerManager.scan(MigrationScannerManager.java:143)
        at org.flywaydb.nc.migration.MigrationScannerManager.lambda$scan$0(MigrationScannerManager.java:64)
        at org.flywaydb.nc.migration.MigrationScannerManager.scan(MigrationScannerManager.java:66)
        at org.flywaydb.nc.utils.VerbUtils.scanForResources(VerbUtils.java:68)
        at org.flywaydb.nc.preparation.PreparationContext.lambda$initialize$0(PreparationContext.java:71)
        at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1789)
```

## Root cause analysis
[ClasspathSqlMigrationScanner#matchesPath](https://github.com/flyway/flyway/blob/502c532b124630f42ffef17fb33755bf5726afe0/flyway-nc/flyway-nc-scanners/src/main/java/org/flywaydb/scanners/ClasspathSqlMigrationScanner.java#L111) (flyway-nc-scanners 12.0.0) dereferences Thread.currentThread().getContextClassLoader().getResource(".") without a null check:

```java
   final String rootPath = new File(Thread.currentThread()
      .getContextClassLoader()
      .getResource(".")              // null under several classloaders
      .getPath()).getAbsolutePath();
```

  ClassLoader.getResource(".") is [not contractual](https://bugs.openjdk.org/browse/JDK-8202687). It returns null under QuarkusClassLoader, OSGi framework loaders (Equinox / Felix), and JDK module-layer loaders, and is undefined on the JDK system loader for multi-entry classpaths (returns only the first entry).

  This reaches my extension via flyway-database-nc-mongodb because the NC pipeline (MigrationScannerManager.scan in flyway-nc-core) runs every registered NativeConnectorsMigrationScanner without consulting Configuration.getResourceProvider(), unlike FlywayExecutor.createResourceAndClassProviders in the classic pipeline, which lets embedders bypass scanning entirely. More details can be found in [my quarkus PR](https://github.com/quarkusio/quarkus/pull/54308#discussion_r3271367936) if needed.

### Option A: Just fix the NPE
`rootPath` is only used in `ClasspathSqlMigrationScanner#matchesPath` to compute `remainingPath = path.substring(rootPath.length() + 1)`. The caller `BaseSqlMigrationScanner.scanFromFileSystem(File dir, Location, …)` already has the on-disk root for that Location. The solution threads `dir` through to a new overload of `matchesPath` that derives the relative path from it, without changing the existing 2-arg abstract method — so the fix is strictly binary- and source-compatible.

  `org.flywaydb.scanners.BaseSqlMigrationScanner`:

  ```java
  // UNCHANGED — still the abstract entry point. Existing subclasses keep
  // compiling and linking against the old class file with no source change.
  abstract boolean matchesPath(String path, Location location);

  // NEW — non-abstract 3-arg overload. Default delegates to the legacy
  // form, so subclasses that don't override it inherit today's behaviour.
  boolean matchesPath(String path, Location location, File rootDir) {
      return matchesPath(path, location);
  }

  // scanFromFileSystem now calls the 3-arg form, passing dir through:
  protected Collection<Pair<LoadableResource, SqlScriptMetadata>> scanFromFileSystem(
          final File dir, final Location location, …) {
      …
      return resourceNames.stream()
          .filter(path -> matchesPath(path, location, dir))   // was matchesPath(path, location)
          …
  }

  org.flywaydb.scanners.ClasspathSqlMigrationScanner:

  // Override the new 3-arg form. The legacy 2-arg body stays as-is but
  // is no longer reached from Flyway's own scanning pipeline, so its
  // NPE is dead code unless a third-party caller invokes it directly.
  @Override
  boolean matchesPath(final String path, final Location location, final File rootDir) {
      final Path relative = rootDir.toPath().relativize(Paths.get(path));
      final String classpathRelative = (location.getRootPath() + "/" + relative)
          .replace("\\", "/");
      return matchesAnyWildcardRestrictions(location, classpathRelative);
  }
```

  FileSystemSqlMigrationScanner is unaffected — it already doesn't call getResource("."), and it inherits the new 3-arg default which delegates back to its existing 2-arg body. It can adopt the 3-arg override later for symmetry, but that's optional.

  ~12 lines, strict semantic improvement on every classloader, no behavioural change on the JDK-loader / single-classpath-entry happy path (the unchanged 2-arg body still runs for any scanner that doesn't override the 3-arg form). The descriptor, visibility, and abstractness of the
  existing matchesPath(String, Location) are unchanged, so binary and source compatibility are preserved for any subclass of BaseSqlMigrationScanner against the old jar.

### Option B: Harmonise flyway-nc with it jdbs version by honouring ResourceProvider in the NC pipeline

  Option A fixes the symptom. The deeper asymmetry is that MigrationScannerManager.scan ignores Configuration.getResourceProvider() entirely, even though FlywayExecutor.createResourceAndClassProviders in the classic pipeline already honours it. Adding the same gate would let embedders skip NC classpath scanning the way they can in the classic pipeline:

```java
  // MigrationScannerManager.scan(...)
  public Collection<LoadableResourceMetadata> scan(
          final Configuration configuration,
          final ParsingContext parsingContext,
          final Location[] locations) {
      final ResourceProvider provider = configuration.getResourceProvider();
      if (provider != null) {
          return resourcesFromProvider(provider, configuration, parsingContext, locations);
      }
      // existing per-scanner flat-map …
  }
```

  Pros: brings NC behaviour in line with the classic pipeline; fixes the same class of bug for all embedders on any non-conforming classloader (Quarkus, OSGi, modular JDK runners, custom enterprise loaders) without touching matchesPath at all; would let the Quarkus MongoDB extension reuse the JDBC-side QuarkusFlywayResourceProvider verbatim.

## Help offered
I'm happy to open a PR for A, B, or both. Just let me know.

seonwooj0810:
Fixes #4241

### Root cause
`ClasspathSqlMigrationScanner.matchesPath` resolved the classpath root with
`Thread.currentThread().getContextClassLoader().getResource(".").getPath()`. Some class
loaders — notably Quarkus' runtime class loader — return `null` for `getResource(".")`, so
`.getPath()` threw a `NullPointerException`, which Flyway surfaced as
`FlywayException: Cannot invoke "java.net.URL.getPath()" because the return value of "java.lang.ClassLoader.getResource(String)" is null`
and aborted application startup.

### Change
The resolved root path is only used to strip the classpath prefix from a discovered
resource before applying a location's wildcard restrictions. Therefore:

- When the location has no path regex (the common case, e.g. `classpath:db/migration`),
  `matchesAnyWildcardRestrictions` returns `true` regardless of the path, so the scanner now
  returns early without resolving the root at all. This mirrors the sibling
  `FileSystemSqlMigrationScanner.matchesPath`.
- When a path regex is present but the classpath root cannot be resolved (`getResource(".")`
  is `null`), it falls back to matching the full normalized path instead of throwing.

Behaviour is unchanged whenever `getResource(".")` resolves normally.

### Verification done
- Verification 1/3/4: confirmed on `main` that `matchesPath` dereferences `getResource(".")`
  unconditionally and that the symptom reproduces; the fix is confined to one `.java` file.
- `./mvnw -pl flyway-nc/flyway-nc-scanners -am compile` succeeds.
- Reproduced and fixed with a standalone harness using a class loader whose `getResource`
  returns `null`: the previous logic throws the exact NPE above, while the patched
  `matchesPath` returns `true` for a plain `classpath:db/migration` location and no longer
  throws for a wildcard-restricted location.

Note: the public repository does not contain a `src/test/java` tree for any module (verified:
no test sources under `flyway-nc/*`), so I could not add an in-repo regression test without
introducing test scaffolding to a module that has none. Happy to add one if you can point me
at where scanner tests live, or adjust the approach for the wildcard-restricted fallback.

### DIFF
--- a/flyway-nc/flyway-nc-scanners/src/main/java/org/flywaydb/scanners/ClasspathSqlMigrationScanner.java
+++ b/flyway-nc/flyway-nc-scanners/src/main/java/org/flywaydb/scanners/ClasspathSqlMigrationScanner.java
@@ -109,12 +109,23 @@ public class ClasspathSqlMigrationScanner extends BaseSqlMigrationScanner {
 
     @Override
     boolean matchesPath(final String path, final Location location) {
-        final String rootPath = new File(Thread.currentThread()
-            .getContextClassLoader()
-            .getResource(".")
-            .getPath()).getAbsolutePath();
-        final String remainingPath = path.replace("\\", "/").substring(rootPath.length() + 1);
+        if (location.getPathRegex() == null) {
+            // Without wildcard restrictions the path is irrelevant. Returning early also avoids
+            // resolving the classpath root, which is not possible under every class loader.
+            return true;
+        }
 
+        final String normalizedPath = path.replace("\\", "/");
+        final URL rootUrl = Thread.currentThread().getContextClassLoader().getResource(".");
+        if (rootUrl == null) {
+            // Some class loaders (e.g. Quarkus) return null for getResource("."), so the classpath
+            // root cannot be determined this way. Fall back to matching the full path instead of
+            // failing with a NullPointerException.
+            return matchesAnyWildcardRestrictions(location, normalizedPath);
+        }
+
+        final String rootPath = new File(rootUrl.getPath()).getAbsolutePath();
+        final String remainingPath = normalizedPath.substring(rootPath.length() + 1);
         return matchesAnyWildcardRestrictions(location, remainingPath);
     }
 


ozgliderpilot:
Hey, @seonwooj0810 .
Thank you for the PR. Did you test this with the QuarkusClassLoader?

seonwooj0810:
Hi @ozgliderpilot — thanks for taking a look.

Honest answer: I haven't run it inside a live Quarkus runtime. The `flyway-nc-scanners` module here doesn't ship a test tree I could exercise `QuarkusClassLoader` against, and I don't have a Quarkus app wired up to reproduce, so I verified the logic against the documented behavior in your issue (and JDK-8202687) rather than end-to-end under Quarkus.

The change targets the two ways your scenario reaches this line:

- **No wildcard location (the common case):** when `location.getPathRegex() == null` the method now returns early and never resolves the classpath root, so `getResource(".")` is never called — that path can't NPE under any class loader.
- **Wildcard location:** if the root is still needed and `getResource(".")` returns `null` (QuarkusClassLoader, per your trace), it falls back to matching the full normalized path via `matchesAnyWildcardRestrictions` instead of dereferencing `null`.

Since you have the actual flyway-mongodb/Quarkus setup that reproduces this, could you confirm it clears the NPE there? And if your extension uses wildcard locations and the full-path fallback doesn't match the way you'd expect, I'm happy to switch to threading the on-disk `dir` through to a relative-path overload (your Option A from the issue) instead — that preserves the relative-path semantics under every class loader. Just let me know which you'd prefer.

ozgliderpilot:
@seonwooj0810, a bit of context of the issue I opened:

The Quarkus `flyway-mongodb` extension currently rewrites Flyway's bytecode at build time (a `BytecodeTransformerBuildItem` that replaces `ClasspathSqlMigrationScanner`) plus a GraalVM `@Substitute` on the read path. These workarounds exist so migrations can be found and read in a packaged app — especially in native mode. The goal is to delete those workarounds and rely on Flyway directly. This PR (the `matchesPath` fix) looked like it might be enough, so I tested it:

- Built Flyway `12.0.0` with this PR's `matchesPath` change applied, installed to the local Maven repo.
- In the extension, removed the two workarounds:
  - the `replaceClasspathScanner` bytecode transform (affects dev + JVM + native),
  - the `FileReadingWithPlaceholderReplacement` GraalVM substitution (native only).
- Ran the extension's migration end-to-end in three modes against a real MongoDB: dev/exploded JVM, fast-jar (packaged `quarkus-run.jar`), and native (GraalVM/Mandrel).

The results:

| Mode | Starts without error? | Migration actually runs? | Where it breaks |
|------|----------------------|--------------------------|-----------------|
| Dev / exploded JVM | yes | **yes** | — works |
| Fast-jar (JVM) | yes (scan finds them) | **no** | **Read** path: `NonJdbcReader` → `FileReadingWithPlaceholderReplacement.readFile` calls `Files.newBufferedReader(Path.of(resource.getAbsolutePath()))`. For a jar-packaged migration the resource is a `ClassPathResource` whose `getAbsolutePath()` is a *relative* classpath path, so `Path.of(...)` resolves against the process CWD → `NoSuchFileException`. App fails to start. |
| Native | yes | **no (silent)** | **Scan** path: `classLoader.getResources("db/migration")` returns a GraalVM `resource:` URL. `ClasspathSqlMigrationScanner.scan()` only handles `file:` and `jar:`, so it skips the URL → 0 migrations discovered → `matchesPath` is never reached. App boots and reports success while migrating nothing. |

So this PR fixes scanning for **dev/exploded JVM only**. Two separate things still block the packaged cases:

1. **Scan** — `ClasspathSqlMigrationScanner.scan()` needs to handle the GraalVM `resource:` protocol (native).
2. **Read** — `FileReadingWithPlaceholderReplacement` / `NonJdbcReader` should read via the resource's stream / classloader (e.g. `LoadableResource.read()` / `getResourceAsStream`) instead of `Files.newBufferedReader(Path.of(getAbsolutePath()))` (fast-jar **and** native).

One caveat worth flagging: with only this PR, the extension's own integration test *appears* to pass — but that's a false positive. It passes only because the forked test process can still see the migrations as loose files under `target/classes` (a `file:` URL → absolute path). Hiding `target/classes/db` and re-running the same packaged jar reproduces the `NoSuchFileException`.

I already have this whole setup running locally (patched Flyway, the extension with workarounds removed, and all three run modes wired up against MongoDB), so it'd be straightforward for me to put up a PR covering the two fixes above. Happy to do that — would you be open to it?

seonwooj0810:
Thanks for the incredibly thorough investigation, @ozgliderpilot — especially for actually wiring up all three run modes against a real MongoDB. That table is exactly the kind of evidence that makes this easy to reason about.

I fully agree with your framing. This PR is deliberately narrow: it just stops `ClasspathSqlMigrationScanner` from throwing an NPE when `getResource(".")` returns null, which only covers the dev/exploded-JVM scan path. The two issues you've isolated are genuinely separate concerns:

1. **Scan** — `scan()` needs to handle the GraalVM `resource:` protocol (currently only `file:` and `jar:`), otherwise native silently discovers 0 migrations.
2. **Read** — `NonJdbcReader` / `FileReadingWithPlaceholderReplacement` should read through the resource's stream (`LoadableResource.read()` / `getResourceAsStream`) instead of `Files.newBufferedReader(Path.of(getAbsolutePath()))`, which breaks for both fast-jar and native.

Great catch on the integration-test false positive, too — that the forked process still sees migrations as loose files under `target/classes` is a subtle trap, and "hide `target/classes/db` and re-run the packaged jar" is a clean way to expose it.

Please do put up that PR — it'd be very welcome, and you clearly already have the ideal repro setup for it. I'm happy to keep this PR scoped to the scan NPE fix so the two can be reviewed independently.

seonwooj0810:
Gentle ping — the discussion with the reporter wrapped up a couple of weeks ago and the change is small and focused (guarding against an NPE when `getResource(".")` returns null). Would a maintainer be able to take a look when there's a chance? Happy to adjust if anything's needed. Thanks!

seonwooj0810:
Coming back to this now that the discussion with @ozgliderpilot wrapped up a while back — let me know if anything's still needed before this can move forward.

PR #4262, opened 2026-07-20
Set Lombok dependency scope to provided across modules

### LINKED ISSUE #4261: Explicitly scope Lombok as provided in published module POM
Several Flyway modules declare `org.projectlombok:lombok` without an explicit scope in the module POM. The child modules rely on the parent `dependencyManagement` section for the scope declaration, but some dependency resolvers do not fully apply inherited scope and can treat Lombok as compile transitive in consumers. Lombok should have scope explicitly declared in the child modules as `<scope>provided</scope>` directly in published module POMs (starting with flyway-core ) to prevent transitive leakage and resolver inconsistencies.

Without such a fix, users of the various flyway modules may have to explicitly exclude lombok in the flyway delcaration to avoid conflicts with the provided version of lombok within their own application. Example:

```
  <dependency>
    <groupId>org.flywaydb</groupId>
    <artifactId>flyway-core</artifactId>
    <version>${flyway.version}</version>
    <exclusions>
      <exclusion>
        <groupId>org.projectlombok</groupId>
        <artifactId>lombok</artifactId>
      </exclusion>
    </exclusions>
  </dependency>
```

JKomoroski:
Fixes #4261

### DIFF
--- a/flyway-command/flyway-command-mcp/pom.xml
+++ b/flyway-command/flyway-command-mcp/pom.xml
@@ -23,6 +23,7 @@
     <dependency>
       <groupId>org.projectlombok</groupId>
       <artifactId>lombok</artifactId>
+      <scope>provided</scope>
     </dependency>
   </dependencies>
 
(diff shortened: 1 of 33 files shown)


pzygielo:
It seems this tries to fix external one
- https://github.com/basepom/dependency-versions-check-maven-plugin/issues/2

Issue #4266, opened 2026-07-23
Are there technical reasons for Flyway 13 requiring Java 21?


phax:
Especially when Flyway is used in "embedded mode", the switch to Java 21 brings some challenges.
Even the current Spring Boot 4.x branch (and many other libraries) still rely on Java 17 as the baseline. So it would be greatly appreciated if you could re-consider changing the minimum version back to Java 17.

So my question is, if there are any technical requirements to do that, or if that was purely to stay "up-to-date".

Any discussion would be appreciated.

pzygielo:
> the switch to Java 21 brings some challenges. 

What are you talking about?

https://github.com/flyway/flyway/blob/c5e20fc77341d4117e7a291214e23f89c1371516/pom.xml#L955-L958

suggests to me that current intention is to compile for 17.

For the latest release 13.0.0:
```
./org/flywaydb/core/api/output/CleanResult.class: compiled Java class data, version 61.0 (Java SE 17)
```
confirms this to be the case.

https://documentation.red-gate.com/flyway/release-notes-and-older-versions/release-notes-for-flyway-engine - I can't see anything about Java 21 there.

> So my question is, if there are any technical requirements to do that, or if that was purely to stay "up-to-date".

The "was" sounds like something had happened already. Could you clarify what do you mean?

phax:
https://documentation.red-gate.com/flyway/reference/usage/api-java states:

> Note: Starting with Flyway v13, Java 21 will be required.

<img width="907" height="401" alt="Image" src="https://github.com/user-attachments/assets/a4221ac4-f201-4327-bd06-3facae825f36" />

Edit: thanks for cross-checking the code. So indeed it still seems to be Java 17... Then the docs are incorrect.

pzygielo:
> https://documentation.red-gate.com/flyway/reference/usage/api-java states:
> 
> > Note: Starting with Flyway v13, Java 21 will be required.

That explains your concerns indeed. (And now probably you are not alone with that.)

> So indeed it still seems to be Java 17... Then the docs are incorrect.

There might be something hiding behind that statement in reference.
But this is for Flyway Team to provide more details.

JasonLuo-Redgate:
Thanks for reporting this issue.

Moving to `Java 21` was originally planned for v13, but that change was postponed and has been deferred to a later release. We'll update the documentation to reflect the current plan 👍

phax:
Thanks @JasonLuo-Redgate for the confirmation.

PR #4276, opened 2026-08-23
Test PR builds on Java 21 and 26

gregjotau:
## What this changes

- Keeps the current Java 17 compatibility floor covered on Linux, macOS, and Windows.
- Adds Linux builds on Java 21 (the documented Flyway 14 floor) and Java 26 (the current OpenJDK feature release).
- Repairs the public PR build command, whose trailing `-pl` currently makes Maven exit with `Missing argument for option: pl` before it builds any module.
- Gives Maven read-only access to the public `flyway-community-db-support` GitHub Packages repository required by `flyway-commandline`.

This is five jobs rather than a full three-JDK/three-OS cross-product: the compatibility floor remains the cross-platform lane, while the upcoming floor and current JDK run once on Linux.

## Why now

Flyway's documentation already says that [Flyway 14 will require Java 21](https://github.com/flyway/flyway/blob/main/documentation/Reference/Usage/API%20(Java).md#supported-java-versions). Testing 21 before changing `<release>17</release>` separates build/runtime compatibility problems from the eventual bytecode baseline change.

The Java 26 lane provides early feedback about compatibility with the current JDK without changing the supported minimum or published bytecode. When Flyway 14 raises the floor, the cross-platform lane can move from 17 to 21 while retaining the current-JDK Linux lane.

I checked the issue and PR history first. Issue #4266 records that the Java 21 change was postponed from v13 to v14; this PR implements supporting CI rather than reopening or duplicating that decision.

## Validation

I ran the complete 48-module reactor from a clean tree on:

- Temurin 17.0.20.1: `BUILD SUCCESS`
- Temurin 21.0.12.1: `BUILD SUCCESS`
- OpenJDK 26.0.2.1: `BUILD SUCCESS`

Command:

```shell
./mvnw -B clean install --file pom.xml
```

The local Maven settings used the same read-only GitHub Packages repository ID configured here. The workflow YAML also parses successfully. This PR does not change Flyway source code, APIs, dependencies, or the Java 17 bytecode target.


gregjotau:
The team can take over this. I dont' need ownership, I just want stuff done and efficient ;)

